PR #6, opened 2024-11-01
Update README.md

ben-allen:
Revisions in response to October 2024 meeting


ben-allen:
Would love to hear feedback on this revision! @eemeli @sffc

sffc:
Seems consistent with what we discussed; don't see any major changes.

PR #28, opened 2025-06-03
First go at spec text for Amount

jessealama:


spec.emu
```diff
@@ -4,36 +4,289 @@
...
+    <h1>Introduction</h1>
+    <p>An Amount is an object that wraps a digit String, representing a mathematical value, and a precision. One can intuitively understand an Amount as a number that, so to speak, knows its own precision.</p>
+  </emu-intro>
+
+  <emu-clause id="sec-amount-abstract-operations">
+    <h1>Abstract Operations</h1>
+
+    <emu-clause id="sec-amount-countsignificantdigits" type="abstract operation">
+      <h1>CountSignificantDigits(
+        _s_: a digit String
```
ljharb, line 35: i don't think "digit String" is currently defined anywhere

spec.emu
```diff
@@ -4,36 +4,289 @@
...
+        <dd>It computes the number of significant digits in a given String, which is assumed to be a digit string in either decimal or exponential notation.</dd>
+      </dl>
+      <emu-alg>
+        1. <mark>Return 42.</mark>
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-amount-countfractiondigits" type="abstract operation">
+      <h1>CountFractionDigits(
+        _s_: a digit String
```
ljharb, line 49: and here, if it can have a `.` then "digit string" is a confusing name, and if not, then couldn't this AO just be "the length of _s_"?

spec.emu
```diff
@@ -4,36 +4,289 @@
...
+      </dl>
+      <emu-alg>
+        1. <mark>Return 42.</mark>
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-amount-applyroundingmodetopositive" type="abstract operation">
+      <h1>ApplyRoundingModeToPositive(
+        _m_: a positive mathematical value,
+        _roundingMode_: a rounding mode
```
ljharb, line 64: where is "rounding mode" defined? it's not in 262, but maybe it's in another proposal?

package.json
```diff
@@ -14,10 +14,13 @@
...
   "devDependencies": {
-    "@tc39/ecma262-biblio": "^2.1.2775",
+    "@tc39/ecma262-biblio": "^2.1.2895",
     "ecmarkup": "^20.0.0"
   },
   "engines": {
     "node": ">= 12"
+  },
+  "dependencies": {
+    "@tc39/ecma402-biblio": "^2.1.1185"
```
ljharb, line 24: probably should also be a dev dep, but not a big deal

spec.emu
ljharb: You can refer to it by generating a biblio file from Decimal, and then checking in that .json file in this repo, and referencing it in the ecmarkup build. https://github.com/tc39/proposal-array-zip/blob/main/package.json#L8 is an example

eemeli: Could this refer to the [rounding modes](https://tc39.es/ecma402/#table-intl-rounding-modes) defined for Intl.NumberFormat, or should the definition from ECMA-402 be brought into ECMA-262 as a part of this proposal?

spec.emu
jessealama: I've added a definition for this now.

spec.emu
jessealama: I'd be happy to refer to the definition in 402, for now. If Decimal gets added to 262, its five rounding modes will be there. I've just pushed a change in which we handle all ECMA-402 rounding modes (we needed to slightly expand the definition of rounding to handle the two rounding modes that IEEE 754 doesn't define, namely `halfTrunc` and `expand`).

spec.emu
```diff
@@ -4,36 +4,421 @@
...
+        1. Let _roundingMode_ be ? GetOption(_opts_, *"roundingMode"*, ~string~, « *"ceil"*, *"floor"*, *"expand"*, *"trunc"*, *"halfCeil"*, *"halfFloor"*, *"halfExpand"*, *"halfTrunc"*, *"halfEven"* », *"halfEven"*).
+        1. Let _unit_ be ? GetOption(_opts_, *"unit"*, ~string~, ~empty~, *undefined*).
+        1. Let _currency_ be ? GetOption(_opts_, *"currency"*, ~string~, ~empty~, *undefined*).
+        1. Let _significantDigits_ be ? GetOption(_opts_, *"significantDigits"*, ~number~, ~empty~, *undefined*).
+        1. Let _fractionDigits_ be ? GetOption(_opts_, *"fractionDigits"*, ~number~, ~empty~, *undefined*).
+        1. If neither _fractionDigits_ nor _significantDigits_ are *undefined*, throw a *RangeError* exception.
+        1. NOTE: We do not support specifying both fractionDigits and significantDigits, even if they actually agree with each other.
+        1. If _fractionDigits_ is not *undefined, then
+          1. If ℝ(_fractionDigits_) is not an integer, throw a *RangeError* exception.
+          1. If ℝ(_fractionDigits_) < 0, throw a *RangeError* exception.
```
ljharb, line 296:
```suggestion
          1. If ℝ(_fractionDigits_) is not a non-negative integer, throw a *RangeError* exception.
```

spec.emu
```diff
@@ -4,36 +4,421 @@
...
+    <h1>Amount.prototype.with ( _options_ )</h1>
+    <p>This method returns a fresh Amount object whose underlying mathematical value is equal to the mathematical value of the current Amount, but possibly with different precision as specified in the _options_.</p>
+    <p>It performs the following steps when called:</p>
+    <emu-alg>
+      1. Let _O_ be the *this* value.
+      1. Perform ? RequireInternalSlot(_O_, [[AmountData]]).
+      1. If _options_ is not an Object, throw a *TypeError* exception.
+      1. Let _v_ be _O_.[[AmountData]].
+      1. Let _s_ be Number::toString(_v_, 10).
+      1. Return ? Call (%Amount%, _O_, « _s_, _options_ »).
```
ljharb, line 419: call, or construct?

jessealama: Ah, good spot. I mean construct. The `OrdinaryObjectCreate` is what I wanted.

spec.emu
```diff
@@ -4,36 +4,395 @@
...
+        1. If _options_ is not an Object, throw a *TypeError* exception.
+      1. Let _v_ be _O_.[[AmountData]].
+      1. Let _numDigits_ be _O_.[[FractionDigits]].
+      1. Assert: _numDigits_ is a non-negative integer.
+      1. Let _options_ be ? GetOptionsObject(_options_).
+      1. Let _roundingMode_ be ? GetOption(_options_, *"roundingMode"*, ~string~, « *"ceil"*, *"floor"*, *"expand"*, *"trunc"*, *"halfCeil"*, *"halfFloor"*, *"halfExpand"*, *"halfTrunc"*, *"halfEven"* », *"halfEven"*).
+      1. Let _fractionDigits_ be ? GetOption(_options_, *"fractionDigits"*, ~number~, ~empty~, *+0*<sub>𝔽</sub>).
+      1. If _fractionDigits_️ is *NaN*, *+∞*<sub>𝔽</sub> or *-∞*<sub>𝔽</sub>, through a *RangeError* exception.
+      1. Let _s_ be RenderMVWithFractionDigits(_v_, ℝ(_fractionDigits_), _roundingMode_).
+      1. If _O_.[[Unit]] is *undefined*, return _s_.
```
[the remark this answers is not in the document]
jessealama: Suggestion: an option that would emit the bare number (no unit).

spec.emu
```diff
@@ -4,36 +4,395 @@
...
+    <p>This method returns a fresh Amount object whose underlying mathematical value is equal to the mathematical value of the current Amount, but possibly with different precision as specified in the _options_.</p>
+    <p>It performs the following steps when called:</p>
+    <emu-alg>
+      1. Let _O_ be the *this* value.
+      1. Perform ? RequireInternalSlot(_O_, [[AmountData]]).
+      1. If _options_ is not an Object, throw a *TypeError* exception.
+      1. Let _v_ be _O_.[[AmountData]].
+      1. Let _s_ be Number::toString(_v_, 10).
+      1. Return ? Call (%Amount%, _O_, « _s_, _options_ »).
+      1. Return OrdinaryObjectCreate(*"%Amount.prototype%"*, « [[AmountData]], [[SignificantDigits]], [[FractionDigits]], [[Unit]] »).
```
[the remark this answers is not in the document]
jessealama: Oops, fixed.

spec.emu
```diff
@@ -4,36 +4,395 @@
...
+    <li>A list of amendments to be made to ECMA-402.</li>
+  </ul>
+</emu-intro>
+
+<emu-clause id="sec-the-amount-object">
+  <h1>The Amount Object</h1>
+  <emu-intro id="sec-amount-intro">
+    <h1>Introduction</h1>
+    <p>An Amount is an object that wraps a decimal digit String, such as `"123.456"` and `1.2e+5`, representing a mathematical value, and a precision. One can intuitively understand an Amount as a number that, so to speak, knows its own precision.</p>
+    <p>The notation for mathematical values used by Amount are <dfn id="dfn-decimal-digit-string">decimal digit string</dfn>, which are Strings that adhere to the <emu-xref href="#prod-DecimalLiteral">DecimalLiteral</emu-xref> production, excepting *"NaN"*, *"Infinity", and *"-Infinity"*.</p>
```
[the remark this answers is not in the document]
jessealama: I've added a bit about units.

spec.emu
```diff
@@ -4,36 +4,395 @@
...
+          1. Set _significantDigits_ to CountSignificantDigits(_digitStr_).
+        1. Otherwise:
+          1. Set _roundedAmountValue_ be RoundToSignificantDigits(_amountValue_, _significantDigits_, _roundingMode_).
+          1. Let _digitStr_ be the unique decimal string representation of _roundedAmountValue_ without leading zeroes.
+          1. Set _fractionDigits_ to CountFractionDigits(_digitStr_).
+        1. Set _O_.[[AmountData]] to _roundedAmountValue_.
+        1. Set _O_.[[SignificantDigits]] to _significantDigits_.
+        1. Set _O_.[[FractionDigits]] to _fractionDigits_.
+        1. If _unit_ is a String, set _O_.[[Unit]] to toLowercase(StringToCodePoints(_unit_)).
+        1. If _currency_ is a String, set _O_.[[Unit]] to toUppercase(StringToCodePoints(_unit_)).
```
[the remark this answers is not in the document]
jessealama: I've added a new `[[Currency]]` internal slot.

spec.emu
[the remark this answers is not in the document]
jessealama: I've added that now: `a.toString({ numberOnly: true }))`. Name improvements welcome.

spec.emu
```diff
@@ -4,36 +4,395 @@
...
+    <h1>Amount.prototype.with ( _options_ )</h1>
+    <p>This method returns a fresh Amount object whose underlying mathematical value is equal to the mathematical value of the current Amount, but possibly with different precision as specified in the _options_.</p>
+    <p>It performs the following steps when called:</p>
+    <emu-alg>
+      1. Let _O_ be the *this* value.
+      1. Perform ? RequireInternalSlot(_O_, [[AmountData]]).
+      1. If _options_ is not an Object, throw a *TypeError* exception.
+      1. Let _v_ be _O_.[[AmountData]].
+      1. Let _s_ be Number::toString(_v_, 10).
+      1. Return ? Call (%Amount%, _O_, « _s_, _options_ »).
```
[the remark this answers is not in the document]
jessealama: Done

spec.emu
```diff
@@ -275,27 +273,27 @@ location: https://github.com/tc39/proposal-measure/
...
+          1. Set _roundedValue_ be RoundToSignificantDigits(_amountValue_, _significantDigits_, _roundingMode_).
+          1. Let _digitStr_ be the unique decimal string representation of _roundedValue_ without leading zeroes.
           1. Set _fractionDigits_ to CountFractionDigits(_digitStr_).
-        1. Set _O_.[[AmountData]] to _roundedAmountValue_.
+        1. Set _O_.[[Value]] to _roundedValue_.
         1. Set _O_.[[SignificantDigits]] to _significantDigits_.
         1. Set _O_.[[FractionDigits]] to _fractionDigits_.
         1. If _unit_ is a String, set _O_.[[Unit]] to toLowercase(StringToCodePoints(_unit_)).
-        1. If _currency_ is a String, set _O_.[[Unit]] to toUppercase(StringToCodePoints(_unit_)).
+        1. If _currency_ is a String, set _O_.[[Currency]] to toUppercase(StringToCodePoints(_unit_)).
```
eemeli, line 296: We should also validate at least that any string values are not empty, and possibly that they match `/^[\w-]+$/`. For currencies, the validation could even require `/^[A-Z]{3}$/i`.

```suggestion
        1. If _unit_ is a String, set _O_.[[Unit]] to CodePointsToString(toLowercase(StringToCodePoints(_unit_))).
        1. If _currency_ is a String, set _O_.[[Currency]] to CodePointsToString(toUppercase(StringToCodePoints(_currency_))).
```

spec.emu
```diff
@@ -355,20 +353,19 @@ location: https://github.com/tc39/proposal-measure/
...
-      1. Let _fractionDigits_ be ? GetOption(_options_, *"fractionDigits"*, ~number~, ~empty~, *+0*<sub>𝔽</sub>).
-      1. If _fractionDigits_️ is *NaN*, *+∞*<sub>𝔽</sub> or *-∞*<sub>𝔽</sub>, through a *RangeError* exception.
-      1. Let _s_ be RenderMVWithFractionDigits(_v_, ℝ(_fractionDigits_), _roundingMode_).
-      1. If _O_.[[Unit]] is *undefined*, return _s_.
-      1. Otherwise, return the string concatenation of _s_, *"["*, _O_.[[Unit]], and *"]"*.
+      1. Let _s_ be RenderMVWithFractionDigits(_v_, _numDigits_, _roundingMode_).
+      1. If _formatBareNumber_ is *true*, return _s_.
+      1. Assert: Either _O_.[[Unit]] nor _O_.[[Currency]] are not *undefined*.
+      1. If _O_.[[Unit]] is *undefined*, let _marker_ be _O_.[[Currency]] else let _marker_ be _O_.[[Unit]].
+      1. Return the string concatenation of _s_, *"["*, _marker_, and *"]"*.
```
eemeli, line 368: The assertion that at least one of currency or unit is defined is not valid.

```suggestion
      1. If _O_.[[Unit]] is a String, then
        1. Assert: _O_.[[Currency]] is not a String.
        1. Return the string concatenation of _s_, *"["*, _O_.[[Unit]], and *"]"*.
      1. Else if _O_.[[Currency]] is a String, then
        1. Return the string concatenation of _s_, *"["*, _O_.[[Currency]], and *"]"*.
```

spec.emu
```diff
@@ -385,12 +382,24 @@ location: https://github.com/tc39/proposal-measure/
...
-      1. Perform ? RequireInternalSlot(_O_, [[AmountData]]).
-      1. If _options_ is not an Object, throw a *TypeError* exception.
-      1. Let _v_ be _O_.[[AmountData]].
-      1. Let _s_ be Number::toString(_v_, 10).
-      1. Return ? Call (%Amount%, _O_, « _s_, _options_ »).
-      1. Return OrdinaryObjectCreate(*"%Amount.prototype%"*, « [[AmountData]], [[SignificantDigits]], [[FractionDigits]], [[Unit]] »).
+      1. Perform ? RequireInternalSlot(_O_, [[Value]]).
+      1. Let _options_ be ? GetOptionsObject(_options_).
+      1. If ? HasOwnProperty(_options_, *"currency"*), throw a *TypeError* exception.
+      1. If ? HasOwnProperty(_options_, *"unit"*), throw a *TypeError* exception.
```
eemeli, line 388: These need to be conditional on _O_.[[Currency]] or _O_.[[Unit]] being a string. Adding a currency or unit to an Amount that doesn't have either ought to be fine.

spec.emu
```diff
@@ -385,12 +382,24 @@ location: https://github.com/tc39/proposal-measure/
...
+      1. Let _significantDigits_ be ? GetOption(_options_, *"significantDigits"*, ~number~, ~empty~, *undefined*).
+      1. If both _fractionDigits_ and _significantDigits_ are not *undefined*, throw a *RangeError* exception.
+      1. If ℝ(_significantDigits_) is not an integer, throw a *RangeError* exception.
+      1. If ℝ(_significantDigits_) < 1, throw a *RangeError* exception.
+      1. If ℝ(_fractionDigits_) is not an integer, throw a *RangeError* exception.
+      1. If ℝ(_fractionDigits_) < 0, throw a *RangeError* exception.
+      1. Let _N_ be OrdinaryObjectCreate(*"%Amount.prototype%"*, « [[Value]], [[SignificantDigits]], [[FractionDigits]], [[Unit]], [[Currency]] »).
+      1. Set _N_.[[Value]] to _O_.[[Value]].
+      1. Set _N_.[[Unit]] to _O_.[[Unit]].
+      1. Set _N_.[[Currency]] to _N_.[[Currency]].
```
eemeli, line 399:
```suggestion
      1. Set _N_.[[Currency]] to _O_.[[Currency]].
```

spec.emu
```diff
@@ -385,12 +382,24 @@ location: https://github.com/tc39/proposal-measure/
...
+      1. Let _options_ be ? GetOptionsObject(_options_).
+      1. If ? HasOwnProperty(_options_, *"currency"*), throw a *TypeError* exception.
+      1. If ? HasOwnProperty(_options_, *"unit"*), throw a *TypeError* exception.
+      1. Let _fractionDigits_ be ? GetOption(_options_, *"fractionDigits"*, ~number~, ~empty~, *undefined*).
+      1. Let _significantDigits_ be ? GetOption(_options_, *"significantDigits"*, ~number~, ~empty~, *undefined*).
+      1. If both _fractionDigits_ and _significantDigits_ are not *undefined*, throw a *RangeError* exception.
+      1. If ℝ(_significantDigits_) is not an integer, throw a *RangeError* exception.
+      1. If ℝ(_significantDigits_) < 1, throw a *RangeError* exception.
+      1. If ℝ(_fractionDigits_) is not an integer, throw a *RangeError* exception.
+      1. If ℝ(_fractionDigits_) < 0, throw a *RangeError* exception.
```
eemeli, line 395: This set of rules will always throw an error for any combination of _fractionDigits_ and _significantDigits_.

eemeli (changes requested):
Some comments, only looking at the diff since my last review.

spec.emu
jessealama: I think that's what I had in mind? What I'm trying to express is that we can't have both. And if either one is specified, they need to be sensible values.

It's possible that you could provide both but in that case I'd like to check that they're coherent. For instance, if we start with `"1.23`, we could do `fractionDigits: 2` and `significantDigits: 3`, and that's OK, but that requires some checking. Is that what you had in mind?

spec.emu
```diff
@@ -292,8 +292,15 @@ location: https://github.com/tc39/proposal-measure/
         1. Set _O_.[[Value]] to _roundedValue_.
         1. Set _O_.[[SignificantDigits]] to _significantDigits_.
         1. Set _O_.[[FractionDigits]] to _fractionDigits_.
-        1. If _unit_ is a String, set _O_.[[Unit]] to toLowercase(StringToCodePoints(_unit_)).
-        1. If _currency_ is a String, set _O_.[[Currency]] to toUppercase(StringToCodePoints(_unit_)).
+        1. If _unit_ is a String, then
+          1. If _unit_ is *""*, throw a *RangeError* exception.
+          1. If _unit_ contains a non-ASCII-alphanumeric character, throw a *RangeError* exception.
+          1. Set _O_.[[Unit]] to CodePointsToString(toLowercase(StringToCodePoints(_unit_))).
```
eemeli, line 298: The `toLowercase()` usage probably needs a similar Unicode reference as in https://tc39.es/ecma262/#sec-string.prototype.tolowercase. Or could we make this an actual call to `String.prototype.toLowerCase()`?

Same applies to _currency_ a few lines below.

spec.emu
```diff
@@ -292,8 +292,15 @@ location: https://github.com/tc39/proposal-measure/
...
+        1. If _currency_ is a String, then
+          1. If the string length of _currency_ is not 3, throw a *RangeError* exception.
+          1. Let _uppercased_ be CodePointsToString(toUppercase(StringToCodePoints(_currency_))).
+          1. If _currency_ is not string equal to _uppercased_, throw a *RangeError* exception.
```
eemeli, line 302: This is also calling out for a syntax-directed operation, but as above, that could be added later.

spec.emu
```diff
@@ -384,19 +394,22 @@ location: https://github.com/tc39/proposal-measure/
...
+      1. Let _unit_ be ? GetOption(_options_, *"unit"*, ~string~, ~empty~, *undefined*).
+      1. Let _currency_ be ? GetOption(_options_, *"currency"*, ~string~, ~empty~, *undefined*).
       1. Let _fractionDigits_ be ? GetOption(_options_, *"fractionDigits"*, ~number~, ~empty~, *undefined*).
       1. Let _significantDigits_ be ? GetOption(_options_, *"significantDigits"*, ~number~, ~empty~, *undefined*).
-      1. If both _fractionDigits_ and _significantDigits_ are not *undefined*, throw a *RangeError* exception.
-      1. If ℝ(_significantDigits_) is not an integer, throw a *RangeError* exception.
-      1. If ℝ(_significantDigits_) < 1, throw a *RangeError* exception.
-      1. If ℝ(_fractionDigits_) is not an integer, throw a *RangeError* exception.
-      1. If ℝ(_fractionDigits_) < 0, throw a *RangeError* exception.
+      1. If _currency_ is not *undefined* and _unit_ is not *undefined*, throw a *RangeError* exception.
```
eemeli, line 401: Also need to throw if either _currency_ or _unit_ is defined, and _O_.[[Currency]] or _O_.[[Unit]] is defined.

They also need to be validated as in the constructor, so it's probably best to define an abstract operation like GetAmountOptions that's used by both the constructor and `with()` to parse the options.

intl.emu
eemeli, line 1: The Intl changes should presume keep-trailing-zeros as its baseline. With that, PluralRules should need no changes, but ToIntlMathematicalValue() will need work.

It'll also be necessary to change how/when errors are thrown for missing `currency` and `unit` values, as that'll no longer be happening in the constructor but during formatting.

jessealama: Done.

I've also tweaked `Amount.prototype [%Symbol.toPrimitive%] ()`, which is also related to the change to `ToIntlMathematicalValue`.

spec.emu
jessealama: Should be done now. I've added a new AO as suggested.

spec.emu
jessealama: I think this should be resolved now. Can you take a look?

eemeli (changes requested):

spec.emu
```diff
@@ -392,29 +403,47 @@ location: https://github.com/tc39/proposal-measure/
...
+  <emu-clause id="sec-amount-symboltoprimitive">
+    <h1>Amount.prototype [ %Symbol.toPrimitive% ] ( _hint_ )</h1>
+    <p>This method is called by ECMAScript language operators to convert an Amount to a primitive value. The allowed values are *"default"*, *"number"*, and *"string"*. The default is *"number."*</p>
+    <p>It performs the following steps when called:</p>
+    <emu-alg>
+      1. Let _O_ be the the *this* value.
+      1. Perform ? RequireInternalSlot(_O_, [[Value]]).
+      1. Perform ? RequireInternalSlot(_O_, [[FractionDigits]]).
+      1. If _O_ has the [[Unit]] internal slot, throw a *TypeError* exception.
+      1. If _O_ has the [[Currency]] internal slot, throw a *TypeError* exception.
```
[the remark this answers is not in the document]
jessealama: Ah, yes, of course, that's what I intended. Fixed now.

spec.emu
```diff
@@ -392,29 +403,47 @@ location: https://github.com/tc39/proposal-measure/
...
-        1. If ℝ(_significantDigits_) < 1, throw a *RangeError* exception.
-      1. Let _N_ be OrdinaryObjectCreate(*"%Amount.prototype%"*, « [[Value]], [[SignificantDigits]], [[FractionDigits]], [[Unit]], [[Currency]] »).
-      1. Set _N_.[[Value]] to _O_.[[Value]].
-      1. If _unit_ is not *undefined*, set _N_.[[Unit]] to _unit_ else set _N_.[[Unit]] to _O_.[[Unit]].
-      1. If _currency_ is not *undefined*, Set _N_.[[Currency]] to _currency_ else set _N_.[[Currency]] to _O_.[[Currency]].
+        1. Set _value_ to RoundToSignificantDigits(_value_, _significantDigits_).
+      1. Let _N_ be OrdinaryObjectCreate(*"%Amount.prototype%"*, « [[Currency]], [[FractionDigits]], [[SignificantDigits]], [[Unit]], [[Value]] »).
+      1. Set _N_.[[Value]] to _value_.
+      1. Set _N_.[[Unit]] to _unit_.
+      1. Set _N_.[[Currency]] to _currency_.
```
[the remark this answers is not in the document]
jessealama: Fixed

intl.emu
```diff
@@ -97,17 +60,25 @@
...
           1. Set _patterns_ to _patterns_.[[percent]].
         1. Else if _style_ is *"unit"*, then
-          1. <ins>If _numberFormat_.[[Unit]] is *undefined*, throw a *TypeError* exception.</ins>
-          1. Let _unit_ be _numberFormat_.[[Unit]].
+          1. <del>Let _unit_ be _numberFormat_.[[Unit]].</del>
+          1. <ins>If _numberFormat_.[[Unit]] is *undefined* and_x_.[[Unit]] is *undefined*, throw a *TypeError* exception.</ins>
+          1. <ins>If _numberFormat_.[[Unit]] is *undefined*, let _unit_ be _x_.[[Unit]].</ins>
+          1. <ins>Else if _x_.[[Unit]] is *undefined*, let _unit_ be _numberFormat_.[[Unit]].</ins>
+          1. <ins>Else if _x_.[[Unit]] and _numberFormat_.[[Unit]] are equal strings, let _unit_ be _numberFormat_.[[Unit]].</ins>
+          1. <ins>Otherwise, throw a *TypeError* exception.</ins>
```
eemeli, line 68: This seems a bit confusing and vague wrt string comparison; would something like this work better?

```suggestion
          1. Let _unit_ be _numberFormat_.[[Unit]].
          1. <ins>Let _fmtUnit_ be _x_.[[Unit]].</ins>
          1. <ins>If _fmtUnit_ is not *undefined* and SameValueNonNumber(_unit_, _fmtUnit_) is *false*, then</ins>
            1. <ins>If _unit_ is not *undefined*, throw a *TypeError* exception.</ins>
            1. <ins>Set _unit_ to _fmtUnit_.</ins>
          1. <ins>Else if _unit_ is *undefined*, throw a *TypeError* exception.</ins>
```

intl.emu
```diff
@@ -97,17 +60,25 @@
...
           1. Set _patterns_ to _patterns_.[[&lt;_unitDisplay_>]].
         1. Else if _style_ is *"currency"*, then
-          1. <ins>If _numberFormat_.[[Currency]] is *undefined*, throw a *TypeError* exception.</ins>
-          1. Let _currency_ be _numberFormat_.[[Currency]].
+          1. <del>Let _currency_ be _numberFormat_.[[Currency]].</del>
+          1. <ins>If _numberFormat_.[[Currency]] is *undefined* and _x_.[[Currency]] is *undefined, throw a *TypeError* exception.</ins>
+          1. <ins>If _x_.[[Currency]] is *undefined*, let _currency_ be _x_.[[Currency]].</ins>
+          1. <ins>Else if _x_.[[Currency]] is *undefined*, let _currency_ be _numberFormat_.[[Currency]].</ins>
+          1. <ins>Else if _x_.[[Currency]] and _numberFormat_.[[Currency]] are equal strings, let _currency_ be _numberFormat_.[[Currency]].</ins>
+          1. <ins>Otherwise, throw a *TypeError* exception.</ins>
```
eemeli, line 81:
```suggestion
          1. Let _currency_ be _numberFormat_.[[Currency]].
          1. <ins>Let _fmtCurrency_ be _x_.[[Currency]].</ins>
          1. <ins>If _fmtCurrency_ is not *undefined* and SameValueNonNumber(_currency_, _fmtCurrency_) is *false*, then</ins>
            1. <ins>If _currency_ is not *undefined*, throw a *TypeError* exception.</ins>
            1. <ins>Set _currency_ to _fmtCurrency_.</ins>
          1. <ins>Else if _currency_ is *undefined*, throw a *TypeError* exception.</ins>
```

intl.emu
```diff
@@ -215,11 +201,12 @@
         1. <ins>If _unit_ is not *undefined* and IsWellFormedUnitIdentifier(_unit_) is *false*, throw a *RangeError* exception.</ins>
         1. Let _unitDisplay_ be ? GetOption(_options_, *"unitDisplay"*, ~string~, « *"short"*, *"narrow"*, *"long"* », *"short"*).
         1. If _style_ is *"currency"*, then
+          1. <ins>If _currency_ is *undefined*, throw a *RangeError* exception.</ins>
           1. Set _intlObj_.[[Currency]] to the ASCII-uppercase of _currency_.
```
eemeli, line 205: If _currency_ is undefined, we should not throw, but just not set _intlObj_.[[Currency]], as is done with _unit_ below.

spec.emu
```diff
@@ -412,11 +412,12 @@ location: https://github.com/tc39/proposal-measure/
...
-      1. If _fractionDigits_ is not *undefined*, then
-        1. Set _value_ to RoundToFractionDigits(_value_, _fractionDigits_).
-      1. Else if _significantDigits_ is not *undefined*, then
-        1. Set _value_ to RoundToSignificantDigits(_value_, _significantDigits_).
+      1. If _fractionDigits_ is not *undefined*, set _value_ to RoundToFractionDigits(_value_, _fractionDigits_).
+      1. Else if _significantDigits_ is not *undefined*, set _value_ to RoundToSignificantDigits(_value_, _significantDigits_).
+      1. Otherwise, throw a *TypeError* exception.
       1. Let _N_ be OrdinaryObjectCreate(*"%Amount.prototype%"*, « [[Currency]], [[FractionDigits]], [[SignificantDigits]], [[Unit]], [[Value]] »).
+      1. If _unit_ is not *undefined* and _O_.[[Currency]] is not *undefined*, throw a *TypeError* exception.
+      1. If _currency_ is not *undefined* and _O_.[[Unit]] is not *undefined*, throw a *TypeError* exception.
```
eemeli, line 420: Thought, could leave fixing for later: We should allow using the same unit/currency as is already in the Amount.

Also, is it intentional to allow
```js
new Amount(42, { unit: 'meter' }).with({ unit: 'liter' })
```
as this would do?

eemeli (changes requested):
Just a few small things left!

eemeli (approved):
🚀 This is good, pending the inline fix to `with()` below.

PR #34, opened 2025-07-16
Bump to stage 2

### LINKED ISSUE #2: Expand this proposal to include Unit-related math and utility functions
This issue is more of a discussion starter, not a final suggestion.

The use cases for a Measure object stretch far beyond the [currently stated use cases and goals](https://github.com/ben-allen/proposal-measure?tab=readme-ov-file#use-cases-and-goals). Constraining the Measure object/API (we could call it an API, right?) to strictly these use cases greatly limits the potential for a standardized measuring API on the web. It makes sense, given that taking input and handling units and measurements is common in many applications.

I argue that inspiration can be taken from the best current state of the art regarding units and measurements on the web.

Among the libraries I've personally explored, is [UnitMath](https://github.com/ericman314/UnitMath).
The notable features offered by UnitMath and not this proposal are, among others:
- representing units as variables, with which [common operations](https://github.com/ericman314/UnitMath?tab=readme-ov-file#arithmetic) and [conversions](https://github.com/ericman314/UnitMath?tab=readme-ov-file#unit-conversion) can be completed seamlessly
- [unit splitting](https://github.com/ericman314/UnitMath?tab=readme-ov-file#unit-conversion)
- [unit simplification](https://github.com/ericman314/UnitMath?tab=readme-ov-file#simplify)
- [custom units and measurement systems](https://github.com/ericman314/UnitMath?tab=readme-ov-file#extending-unitmath)
- [unit serialization](https://github.com/ericman314/UnitMath?tab=readme-ov-file#formatting)

While not all features offered by this library (or others) should necessarily be included in a standard web API, there is experience in these libraries that could be applied or considered in this proposal to make for a better spec.

I invite @ericman314, the creator of UnitMath, to leave his thoughts here.

### LINKED ISSUE #10: Unit spec: CLDR, UCOM, or something else?
Unicode CLDR gives a list of units: https://github.com/unicode-org/cldr/blob/main/common/supplemental/units.xml

There is also another list of units in a standard called The Unified Code for Units of Measure published by the Regenstrief Institute, which appears to be associated with Indiana University. https://ucum.org/ucum

Of course there is SI itself, published by the BIPM. https://www.bipm.org/en/

These different lists have different features and were made for different use cases. Unicode CLDR is primarily oriented toward units for human display; it is nominally what we use in ECMA-402 (although ECMA-402 contains only a small subset of it), and one key feature is its support for mixed units such as `foot-and-inch`. I am not as familiar with the other systems, so I can't speak as well to their advantages and disadvantages.

CC @zbraniecki @macchiati @roozbehp

### LINKED ISSUE #11: Mixed units: index by largest or smallest unit?
In a mixed unit such as "foot-and-inch", in what unit should the scalar value be indexed?

Should 5 feet 10 inches be:

1. 70 foot-and-inch (indexed in smallest unit)
2. 5.833 foot-and-inch (indexed in largest unit)

ICU internally does option 2, but I don't see that behavior discussed in UTS 35.

@macchiati @younies @ben-allen

### LINKED ISSUE #17: Is the precision with respect to the scalar value, or the formatted value?
With mixed units, we're liable to end up representing e.g. 5ft 10in as something close to 5.83333, which when formatting needs precision to be defined, so that we don't end up with "5ft 9.99996in".

As with the scalar value that's discussed in #11, we need to determine whether the precision would be indexed by the smallest or largest unit:
1. `precision: 2` if by the largest unit
2. `precision: 0` if by the smallest unit

Even if we end up holding the value indexed by the largest unit, I'm starting to think that the precision ought to be measured as fractional digits of the _smallest_ unit. Otherwise, we end up needing to come up with way too many solutions to corner-cases like what `precision: 1` might mean in the above, or what `precision: 2` means when formatting a value like "5ft 1in".

A key realisation here (at least for me) is that the total precision of a measurement should take into account the units: when formatting with `foot-and-inch`, there's a built-in expectation of the value including inches, which isn't present when formatting `foot` units. Therefore, the `precision` is most useful if it's considered to be going beyond the smallest declared unit.

### LINKED ISSUE #20: Add Format API, `Measure.prototype.toString` may not be enough
What about a `format` API interoperable with `Intl` API design ?

see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat
see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

Provide an `Intl.MeasureFormat` API and `Measure.prototype.toLocaleString` shortcut.

### LINKED ISSUE #21: How should be managed unknown units with conversion ?
https://en.wikipedia.org/wiki/Hectare#Are

Let say runtime known `meter` unit, be we want work with `Are` (1a = 100m2).

Should this proposal allow to add custom units and describe possible conversions ?

API Idea (pseudocode):

```js
Measure.addUnit({
  symbol: 'a',
  name: 'are',
  kind: {type: 'distance', exponent: 2} // surface
  conversions: [
    {
      type: 'distance',
      exponent: 2,
      unit: 'meter',
      convertTo: (ares: number) => ares * 100,
      convertFrom: (m2: number) => m2 / 100,
    }
  ],
});

const measureAre = new Measure(1, {unit: 'are'});
const measureM2 = measureAre.convertTo('meter');
// Measure { value: 100, unit: 'meter', exponent: 2}
```

jessealama:


eemeli:
Slides for the upcoming TC39 presentation: https://docs.google.com/presentation/d/1my6X1ODDckzJmtcWcFI9hRF_I06Z4RQwrq81lbo8wPM/edit?usp=sharing

README.md
```diff
@@ -1,6 +1,6 @@
 # Representing Measures
```
eemeli, line 1:
```suggestion
# Representing Amounts
```

eemeli (approved):

PR #44, opened 2025-07-24
Drop currency option

eemeli:
Fixes #36 by dropping the `currency` option. It's left in the Intl mathematical value, as that'll need slightly special treatment, and the Intl side of things needs to be properly rebased on top of the [keep-trailing-zeros](https://github.com/tc39/proposal-intl-keep-trailing-zeros) proposal in any case.

CC @gibson042, @sffc


gibson042 (approved):
Thanks!

PR #67, opened 2025-09-22
Remove redundant step in ComputeFractionDigits SDO

jessealama:
I intended to push this change, as indicated [here](https://github.com/tc39/proposal-amount/pull/63#discussion_r2368091811), but the commit didn't get pushed.


jessealama:
Already merged in.

PR #101, opened 2026-04-23
Use `fractionDigits` & `significantDigits` options in `.convertTo()`, allow Number values

eemeli:
Following today's discussion on https://github.com/tc39/proposal-amount/pull/90#issuecomment-4305702252, this PR:
- Updates the README to match the spec changes applied in #89.
- Replaces the `.convertTo()` precision options (minimum/maximum fraction/significant digits) with those of the constructor, i.e. `fractionDigits` and `significantDigits`, and updates their handling to correspond to what we're already doing in the constructor. `roundingPriority` is dropped, as specifying both `fractionDigits` and `significantDigits` is an error.
- Changes it so that if you `.convertTo()` without any precision options, you end up with a Number-backed Amount.

CC @sffc, @gibson042

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ The object prototype would provide the following methods:
   Otherwise, the returned Amount will have a Number `value`.
   If both `fractionDigits` and `significantDigits` are set, a RangeError is thrown.
 
-  Calling `convertTo()` will throw a RangeError if conversion is not supported
+  Calling `convertTo()` will throw a TypeError if conversion is not supported
   for the Amount's unit (such as currency units),
   or if the resolved conversion target is not valid for the Amount's unit
   (such as attempting to convert a mass unit into a length unit).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,29 +90,24 @@ The object prototype would provide the following methods:
...
-  stored as a String (if finite) or Number (if not finite).
+  If the result of unit conversion is finite and either the
+  `fractionDigits` or `significantDigits` options are set,
+  the result will be rounded according to the precision options,
+  and the returned Amount will have a String `value`.
+  Otherwise, the returned Amount will have a Number `value`.
+  If both `fractionDigits` and `significantDigits` are set, a RangeError is thrown.
 
-  Calling `convertTo()` will throw an error if conversion is not supported
+  Calling `convertTo()` will throw a RangeError if conversion is not supported
```
gibson042: This should be a TypeError (cf. https://github.com/tc39/proposal-amount/pull/89#discussion_r3024844847 and [#92](https://github.com/tc39/proposal-amount/issues/92#:~:text=replacement%20of%20RangeError%20with%20TypeError%20to%20indicate%20conflicting%20options%2C%20aligning%20with%20e.g.%20CreateDateTimeFormat%20and%20SetNumberFormatDigitOptions))
```suggestion
  Calling `convertTo()` will throw a TypeError if conversion is not supported
```

gibson042 (approved):

jessealama (approved):

Issue #95, opened 2026-04-02
ECMA-402 layering is incorrect


gibson042:
It will be necessary for ECMA-402 operations to supersede those in ECMA-262, but the current spec text handles that incorrectly—ECMA-402 requires GetAmountConvertToOptions to provide locale and usage by reading properties from _options_ [_[at the right time](https://github.com/tc39/proposal-amount/issues/92)_], while ECMA-262 requires absence of such [[Get]] operations. This proposal currently defines a [GetAmountConvertToOptions](https://tc39.es/proposal-amount/#sec-amount-getamountconverttooptions) that is documented to return a Record with fields [[Locale]] and [[Usage]], but fails to do so.

There's also what looks to me like an uncomfortable about of redundancy between the [ECMA-262 Amount.prototype.convertTo](https://tc39.es/proposal-amount/#sec-amount.prototype.convertto) and the [ECMA-402 Amount.prototype.convertTo](https://tc39.es/proposal-amount/#sup-amount.prototype.convertto) which makes it difficult to observe what is actually different.

I think things would be well-structured like so:
* There is a single common GetAmountConvertToOptions, with the reading and validation of `locale`, `localeMatcher`, and `usage` required by ECMA-402 but absent in ECMA-262 being handled as either normative optional steps (analogous to [ToBoolean](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-toboolean)) or as calls to operations with no-op default implementations and superseding definitions in ECMA-402 (analogous to [ECMA-262 Array.prototype.toLocaleString](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.tolocalestring) and its [ECMA-402 replacement](https://tc39.es/ecma402/#sup-array.prototype.tolocalestring)).
* The majority of Amount.prototype.convertTo is likewise common to both specs, delegating to an operation like Let _targetUnit_ be ? SelectUnitConversionTarget(_sourceValue_, _sourceUnit_, _validatedOpts_) with definitions like the following:
  * ECMA-262:
    ```
    1. Let _unit_ be _validatedOpts_.[[Unit]].
    1. If _unit_ is *undefined*, throw a *TypeError* exception.
    1. Return _unit_.
    ```
  * ECMA-402:
    ```
    1. Let _unit_ be _validatedOpts_.[[Unit]].
    1. Let _requestedLocales_ be _validatedOpts_.[[RequestedLocales]].
[...]

PR #110, opened 2026-05-16
Add GetAmountConvertToUnit definitions to both ECMA-262 and ECMA-402

### LINKED ISSUE #109: Add spec text for handling for `locale` and `usage` conversion options
As noted by @waldemarhorwat in #108, we're missing an AO in the proposed spec changes that'll handle the `locale` and `usage` conversion options. It'll probably be an AO called from [GetAmountConvertToOptions](https://tc39.es/proposal-amount/#sec-amount-getamountconverttooptions) that'll be overridden in ECMA-402 to look up the data from the CLDR `units.xml`.

eemeli:
Fixes #109 by adding an AO to ECMA-262 that's superseded by a redefinition in ECMA-402.

~The actual extraction of preferred units from the CLDR `units.xml` is left as a TODO.~ Edit: now included.

The preferences depend on unit category, usage, region, and value thresholds. The categories we get from [validity/unit.xml](https://github.com/unicode-org/cldr/blob/main/common/validity/unit.xml), the region is calculated from the first-choice locale, and the preferences themselves are in [supplemental/units.xml](https://github.com/unicode-org/cldr/blob/main/common/supplemental/units.xml).

As the data always includes `001` ("the world") as one of the regions for each supported category + usage combination, no locale fallback is ever done during `.convertTo()`, as all well-formed locales are supported.


spec.emu
```diff
@@ -98,26 +98,27 @@ location: https://github.com/tc39/proposal-amount/
             1. If _significantDigits_ is not an integral Number, throw a *RangeError* exception.
             1. Set _significantDigits_ to ℝ(_significantDigits_).
             1. If _significantDigits_ is not in the inclusive interval from 1 to 21, throw a *RangeError* exception.
-          1. If _unit_ is the empty String, throw a *RangeError* exception.
```
jessealama: Is the empty string allowed now because unit conversion potentially succeeds? Or perhaps we're taking the stance that all strings are, in principle, acceptable?

jessealama:
This looks good. Thanks!

jessealama (approved):

spec.emu
eemeli: Ach, this is accidental; I got lost between the constructor and conversion options handling.

spec.emu
```diff
@@ -98,26 +98,27 @@ location: https://github.com/tc39/proposal-amount/
             1. If _significantDigits_ is not an integral Number, throw a *RangeError* exception.
             1. Set _significantDigits_ to ℝ(_significantDigits_).
             1. If _significantDigits_ is not in the inclusive interval from 1 to 21, throw a *RangeError* exception.
```
eemeli, line 125:
```suggestion
            1. If _significantDigits_ is not in the inclusive interval from 1 to 21, throw a *RangeError* exception.
          1. If _unit_ is the empty String, throw a *RangeError* exception.
```

spec.emu
```diff
@@ -105,19 +105,21 @@ location: https://github.com/tc39/proposal-amount/
...
+          <dd>It validates the given _options_ for converting an Amount to another Amount and returns a Record with slots set to appropriate marthematical values (or *undefined*).</dd>
         </dl>
         <emu-alg>
-          1. Let _opts_ be ? GetOptionsObject(_opts_).
+          1. Let _opts_ be ? GetOptionsObject(_options_).
           1. Let _fractionDigits_ be ? GetOption(_opts_, *"fractionDigits"*, ~number~, ~empty~, *undefined*).
           1. Let _roundingMode_ be ? GetOption(_opts_, *"roundingMode"*, ~string~, « *"ceil"*, *"floor"*, *"expand"*, *"trunc"*, *"halfCeil"*, *"halfFloor"*, *"halfExpand"*, *"halfTrunc"*, *"halfEven"* », *"halfEven"*).
           1. Let _significantDigits_ be ? GetOption(_opts_, *"significantDigits"*, ~number~, ~empty~, *undefined*).
-          1. Let _unit_ be ? GetOption(_opts_, *"unit"*, ~string~, ~empty~, *undefined*).
+          1. Let _unit_ be ? GetAmountConvertToUnit(_sourceValue_, _sourceUnit_, _opts_).
```
gibson042: This change makes `amount.convertTo({ fractionDigits: 1.5, unit: "" })` throw because of `unit` while `new Amount(0, { fractionDigits: 1.5, unit: "" })` throws because of `fractionDigits`. I think the new behavior is better (#92), but more importantly, the algorithms should be kept in sync. Ideally, they just get combined into a single operation for Amount to call like `? GetAmountOptions(_opts_)` and Amount.prototype.convertTo to call like `? GetAmountOptions(_opts_, _sourceValue_, _sourceUnit_)`, with steps like
```diff
           1. Let _significantDigits_ be ? GetOption(_opts_, *"significantDigits"*, ~number~, ~empty~, *undefined*).
-          1. Let _unit_ be ? GetAmountConvertToUnit(_sourceValue_, _sourceUnit_, _opts_).
+          1. Let _unit_ be ? GetOption(_opts_, *"unit"*, ~string~, ~empty~, *undefined*).
+          1. If _unit_ is the empty String, throw a *RangeError* exception.
+          1. If _sourceValue_ is present or _sourceUnit_ is present, then
+            1. Assert: _sourceValue_ is present and _sourceUnit_ is present.
+            1. Set _unit_ to ? SelectUnitConversionTarget(_sourceValue_, _sourceUnit_, _unit_, _opts_).
```

intl.emu
```diff
@@ -11,6 +11,134 @@
...
+    <h1>Abstract Operations of the Amount Object</h1>
+
+    <emu-clause id="sec-amount-unit-category-data">
+      <h1>Unit Category Data</h1>
+      <p>
+        Unit category data is derived from CLDR file
+        <a href="https://github.com/unicode-org/cldr/blob/main/common/validity/unit.xml"><code>validity/unit.xml</code></a>.
+        As described in <a href="https://unicode.org/reports/tr35/tr35.html#Validity_Data">Unicode Technical Standard #35 Part 1, Validity Data</a>,
+        the body of the <code>&lt;id type='unit' idStatus='regular'></code> element contains
+        the supported units, represented as a whitespace-delimited list of sequences of subtags.
```
gibson042:
```suggestion
        the supported units, represented as a whitespace-separated list in which each element is a sequence of subtags separated by *"-"* (U+002D HYPHEN-MINUS) characters.
```

intl.emu
```diff
@@ -11,6 +11,134 @@
...
 
+  <emu-clause id="sup-amount-abstract-operations">
+    <h1>Abstract Operations of the Amount Object</h1>
+
+    <emu-clause id="sec-amount-unit-category-data">
+      <h1>Unit Category Data</h1>
+      <p>
+        Unit category data is derived from CLDR file
+        <a href="https://github.com/unicode-org/cldr/blob/main/common/validity/unit.xml"><code>validity/unit.xml</code></a>.
+        As described in <a href="https://unicode.org/reports/tr35/tr35.html#Validity_Data">Unicode Technical Standard #35 Part 1, Validity Data</a>,
```
gibson042:
```suggestion
        As described in <a href="https://unicode.org/reports/tr35/tr35.html#Validity_Data">Unicode Technical Standard #35 Part 1 Core, Validity Data</a>,
```

intl.emu
```diff
@@ -11,6 +11,134 @@
...
+        the body of the <code>&lt;id type='unit' idStatus='regular'></code> element contains
+        the supported units, represented as a whitespace-delimited list of sequences of subtags.
+        For each unit, the first subtag provides the unit's category, and the remaining subtags are the unit identifier.
```
gibson042:
```suggestion
        For each unit, the first subtag provides the <dfn variants="unit categories">unit category</dfn>, and the remaining subtags are the unit identifier.
```

intl.emu
```diff
@@ -11,6 +11,134 @@
...
+      </p>
+    </emu-clause>
+
+    <emu-clause id="sec-amount-unit-preference-data">
+      <h1>Unit Preference Data</h1>
+      <p>Unit preference data is derived from CLDR file <a href="https://github.com/unicode-org/cldr/blob/main/common/supplemental/units.xml"><code>supplemental/units.xml</code></a>. As described in <a href="https://unicode.org/reports/tr35/tr35-info.html#Unit_Preferences_Data">Unicode Technical Standard #35 Part 6 Supplemental, Unit Preferences Data</a>, each <code>&lt;unitPreferences&gt;</code> element contains the preferred units for its <code>category</code> and <code>usage</code>. A <code>&lt;unitPreferences&gt;</code> element contains one or more <code>&lt;unitPreference&gt;</code> elements, which define the <code>regions</code> (a space-delimited list of region identifiers) and optionally the <code>geq</code> threshold for the unit preference.</p>
```
gibson042: Vocabulary nit:
```suggestion
      <p>Unit preference data is derived from CLDR file <a href="https://github.com/unicode-org/cldr/blob/main/common/supplemental/units.xml"><code>supplemental/units.xml</code></a>. As described in <a href="https://unicode.org/reports/tr35/tr35-info.html#Unit_Preferences_Data">Unicode Technical Standard #35 Part 6 Supplemental, Unit Preferences Data</a>, each <code>&lt;unitPreferences&gt;</code> element contains the preferred units for its <code>category</code> and <code>usage</code>. A <code>&lt;unitPreferences&gt;</code> element contains one or more <code>&lt;unitPreference&gt;</code> elements, which define the <code>regions</code> (a space-separated list of region identifiers) and optionally the <code>geq</code> threshold for the unit preference.</p>
```

intl.emu
```diff
@@ -11,6 +11,134 @@
...
+      <p>This definition supersedes the definition provided in <emu-xref href="#sec-amount-getamountconverttounit"></emu-xref>.</p>
+      <emu-alg>
+        1. Let _unit_ be ? GetOption(_options_, *"unit"*, ~string~, ~empty~, *undefined*).
+        1. Let _usage_ be ? GetOption(_options_, *"usage"*, ~string~, ~empty~, *undefined*).
+        1. Let _locale_ be ? Get(_options_, *"locale"*).
+        1. If _unit_ is a String, then
+          1. If _usage_ is not *undefined*, throw a *TypeError* exception.
+          1. If _locale_ is not *undefined*, throw a *TypeError* exception.
+          1. If _unit_ is the empty String, throw a *RangeError* exception.
+          1. Return _unit_.
```
gibson042:
```suggestion
        1. Let _unit_ be ? GetOption(_options_, *"unit"*, ~string~, ~empty~, *undefined*).
        1. If _unit_ is the empty String, throw a *RangeError* exception.
        1. Let _usage_ be ? GetOption(_options_, *"usage"*, ~string~, ~empty~, *undefined*).
        1. Let _locale_ be ? Get(_options_, *"locale"*).
        1. If _unit_ is a String, then
          1. If _usage_ is not *undefined* or _locale_ is not *undefined*, throw a *TypeError* exception.
          1. Return _unit_.
```
(I also have comments on the ordering at #92, but this at least localizes validation of `unit`)

intl.emu
```diff
@@ -11,6 +11,134 @@
...
+        1. Let _maximalLocale_ be the result of the <a href="https://unicode.org/reports/tr35/#Likely_Subtags">Add Likely Subtags</a> algorithm applied to _resolvedLocale_. If an error is signaled, set _maximalLocale_ to _resolvedLocale_.
+        1. Let _region_ be GetLocaleRegion(_maximalLocale_).
+        1. If _region_ is *undefined*, set _region_ to *"001"*.
+        1. Let _preferredUnits_ be ? GetPreferredUnits(_category_, _usage_, _region_).
+        1. Repeat, while _preferredUnits_ has more than one element,
+          1. Let _entry_ be the first element of _preferredUnits_.
+          1. Set _unit_ to _entry_.[[Unit]].
+          1. Let _threshold_ be _entry_.[[Threshold]].
+          1. Let _convertedValue_ be ? ConvertUnitValue(_sourceValue_, _sourceUnit_, _unit_).
+          1. If _convertedValue_ is not finite or _convertedValue_ ≥ _threshold_, return _unit_.
```
gibson042: These steps seem to have some problems:

* _threshold_ is a mathematical value but _convertedValue_ is a Number
* per [UTS #35 Part 6 Supplemental, Unit Preferences Data](https://unicode.org/reports/tr35/tr35-info.html#Unit_Preferences_Data), `geq` must be compared against an absolute value

```suggestion
          1. Let _convertedValue_ be ? ConvertUnitValue(_sourceValue_, _sourceUnit_, _unit_).
          1. If _convertedValue_ is not finite or abs(ℝ(_convertedValue_)) ≥ _threshold_, return _unit_.
```

intl.emu
```diff
@@ -11,6 +11,134 @@
...
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It determines the preferred units to use for the given category, usage, and region.</dd>
+      </dl>
+      <emu-alg>
+        1. If _category_ is the <code>category</code> and _usage_ is the <code>usage</code> of a <code>&lt;unitPreferences&gt;</code> element in the <emu-xref href="#sec-amount-unit-preference-data">unit preference data</emu-xref>, then
+          1. Let _unitPreferencesParent_ be that element.
+        1. Else,
+          1. Throw a *RangeError* exception.
```
gibson042: These steps seem to be missing some important functionality from [Compute the preferred output unit](https://unicode.org/reports/tr35/tr35-info.html#compute-the-preferred-output-unit), e.g. "land-agriculture-grain ⊂ land-agriculture ⊂ land ⊂ default" and "CH (Switzerland) ⊂ 155 (Western Europe) ⊂ 150 (Europe) ⊂ 001 (World)".

intl.emu
```diff
@@ -11,6 +11,134 @@
...
+        1. Assert: _region_ can be matched by the <code>unicode_region_id</code> Unicode locale nonterminal.
```
gibson042: This seems like it should be covered by a more specific type for the parameter.

intl.emu
```diff
@@ -11,6 +11,134 @@
...
+      <h1>
+        GetPreferredUnits (
+          _category_: a String,
+          _usage_: a String,
+          _region_: a String,
+        ): either a normal completion containing a List of Records or a throw completion
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It determines the preferred units to use for the given category, usage, and region.</dd>
```
gibson042:
```suggestion
        <dd>It determines the preferred units to use for the given category, usage, and region, and returns them ordered by descending threshold.</dd>
```

intl.emu
```diff
@@ -11,6 +11,134 @@
...
+          1. Let _regionsAttr_ be the String value of the <code>regions</code> attribute of _unitPreference_.
+          1. Let _regions_ be StringSplitToList(_regionsAttr_, *" "*).
+          1. Let _geq_ be the mathematical value of the <code>geq</code> attribute of _unitPreference_ if present, or 1 otherwise.
+          1. Let _unit_ be the String value of the body of the _unitPreference_ element.
+          1. If _regions_ contains _region_, append the Record { [[Unit]]: _unit_, [[Threshold]]: _geq_ } to _match_.
+          1. Else if _regions_ contains *"001"*, append the Record { [[Unit]]: _unit_, [[Threshold]]: _geq_ } to _fallback_.
+        1. If _match_ is not empty, return _match_.
+        1. Assert: _fallback_ is not empty.
+        1. Return _fallback_.
+      </emu-alg>
```
gibson042:
```suggestion
      </emu-alg>
      <emu-note>
        <p><a href="https://unicode.org/reports/tr35/tr35-info.html#constraints">Unicode Technical Standard #35 Part 1, Unit Preferences Constraints</a> specifies that the <emu-xref href="#sec-amount-unit-preference-data">unit preference data</emu-xref> for a given category, usage, and region-set is already in descending order [by threshold].</p>
      </emu-note>
```

gibson042 (changes requested):
Thanks! I have many suggestions, but this is an _excellent_ starting point.

intl.emu
eemeli: I'd missed that, thanks. I'll add the ~category~ **usage** fallback.

On region fallback, I'd prefer to follow the LDML note
> This loop can be optimized to only include containing regions that occur in the data (eg, only 001 in LDML 45).

as the unit data continues not to include any numeric codes other than 001. With that in mind, I'd prefer not introducing this chain of region fallbacks unless that actually becomes necessary.

intl.emu
eemeli: Do you mean something like this?
> _region_: a String matching the <code>unicode_region_id</code> Unicode locale nonterminal

spec.emu
eemeli: In b853e19 I made it so that the handling of the options shared between the constructor and unit conversion is always done via GetAmountOptions.

intl.emu
gibson042: I think it's necessary to match the algorithm defined in UTS #35 now, otherwise if such data _does_ get introduced then implementations using ICU will become nonconformant upon upgrade and implementations not using ICU will provide inferior user experience. We want to align with that upstream standard.

intl.emu
gibson042: Yes, but I see now that even [GetLocaleRegion](https://tc39.es/ecma402/#sec-getlocaleregion) is documented to return just a String, so I'm comfortable with deferring this to a later cleanup.

intl.emu
eemeli: Usage fallbacking is added in 64a9f69.

I'll add a TODO about the region containment fallbacking.

spec.emu
```diff
@@ -105,30 +105,47 @@ location: https://github.com/tc39/proposal-amount/
...
-          1. Else if _significantDigits_ is not *undefined*, then
-            1. If _significantDigits_ is not an integral Number, throw a *RangeError* exception.
-            1. Set _significantDigits_ to ℝ(_significantDigits_).
-            1. If _significantDigits_ is not in the inclusive interval from 1 to 21, throw a *RangeError* exception.
-          1. If _unit_ is the empty String, throw a *RangeError* exception.
-          1. Return the Record { [[FractionDigits]]: _fractionDigits_, [[RoundingMode]]: _roundingMode_, [[SignificantDigits]]: _significantDigits_, [[Unit]]: _unit_ }.
+          1. If _options_ is not an Object, throw a *TypeError* exception.
+          1. Let _resolvedOpts_ be ? GetAmountOptions(_options_).
+          1. Let _unitOption_ be _resolvedOpts_.[[Unit]].
+          1. Let _unit_ be ? GetAmountConvertToUnit(_sourceValue_, _sourceUnit_, _options_, _unitOption_).
```
gibson042: This seems incorrect w.r.t. #92 (_specifically in rejecting an undefined `unit` too late_), but that can be sorted out later. To reïterate, though, I do expect it to end up something like [`GetAmountOptions ( options [ , sourceValue, sourceUnit ] )`](https://github.com/tc39/proposal-amount/pull/110#discussion_r3256797479), with the operation switching into an optional mode where it ensures that the resolved [[Unit]] is not undefined (_via logic that differs between base ECMA-262 and ECMA-402_).

And per #95, I also expect that mode to observably read all options in alphabetical order, including `locale` and `usage` iff the implementation includes ECMA-402.

spec.emu
```diff
@@ -105,30 +105,47 @@ location: https://github.com/tc39/proposal-amount/
...
+          1. Set _resolvedOpts_.[[Unit]] to _unit_.
+          1. Return _resolvedOpts_.
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-amount-getamountconverttounit" type="abstract operation">
+        <h1>
+          GetAmountConvertToUnit (
```
gibson042: Suggestion: rename (both here and in intl.emu).
```suggestion
          SelectTargetUnit (
```

spec.emu
```diff
@@ -105,30 +105,47 @@ location: https://github.com/tc39/proposal-amount/
...
-          _opts_: an Object
+          _sourceValue_: a Number,
+          _sourceUnit_: a String,
+          _options_: an ECMAScript language value,
         ): either a normal completion containing a Record with fields [[FractionDigits]] (a non-negative integer or *undefined*), [[RoundingMode]] (a <emu-xref href="#dfn-amount-rounding-mode">rounding mode</emu-xref>), [[SignificantDigits]] (a positive integer or *undefined*), and [[Unit]] (a String or *undefined*) or a throw completion
         </h1>
         <dl class="header">
           <dt>description</dt>
-          <dd>It validates the given _options_ (an ECMAScript object) for converting an Amount to another Amount and returns a Record with slots set to appropriate marthematical values (or *undefined*).</dd>
+          <dd>It validates the given _options_ for converting an Amount to another Amount and returns a Record with slots set to appropriate marthematical values (or *undefined*).</dd>
```
gibson042:
```suggestion
          <dd>It validates the given _options_ for translating an Amount's value to a new unit and returns their resolved values.</dd>
```

spec.emu
```diff
@@ -76,15 +76,15 @@ location: https://github.com/tc39/proposal-amount/
...
       <emu-clause id="sec-amount-getamountoptions" type="abstract operation">
         <h1>GetAmountOptions (
-          _opts_: an Object
+          _options_: an ECMAScript language value,
         ): either a normal completion containing a Record with fields [[FractionDigits]] (a non-negative integer or *undefined*), [[RoundingMode]] (a <emu-xref href="#dfn-amount-rounding-mode">rounding mode</emu-xref>), [[SignificantDigits]] (a positive integer or *undefined*), and [[Unit]] (a String or *undefined*) or a throw completion
         </h1>
         <dl class="header">
           <dt>description</dt>
-          <dd>It validates the given _options_ (an ECMAScript object) for creating an Amount and returns a Record with slots set to appropriate marthematical values (or *undefined*).</dd>
+          <dd>It validates the given _options_ for creating an Amount and returns a Record with slots set to appropriate marthematical values (or *undefined*).</dd>
```
gibson042: This operation is no longer unique to Amount creation.
```suggestion
          <dd>It validates the given _options_ and returns their resolved values.</dd>
```

intl.emu
```diff
@@ -11,6 +11,172 @@
...
+      <emu-alg>
+        1. Let _usage_ be ? GetOption(_options_, *"usage"*, ~string~, ~empty~, *undefined*).
+        1. Let _locale_ be ? Get(_options_, *"locale"*).
+        1. If _unitOption_ is a String, then
+          1. If _usage_ is not *undefined* or _locale_ is not *undefined*, throw a *TypeError* exception.
+          1. Return _unitOption_.
+        1. If _usage_ is *undefined*, then
+          1. If _locale_ is *undefined*, throw a *TypeError* exception.
+          1. Set _usage_ to *"default"*.
+        1. Let _category_ be ? GetUnitCategory(_sourceUnit_).
```
gibson042, line 59: Since this depends upon the `convertTo` receiver rather than invocation arguments, I think it should be performed before interacting with _options_ at all. But again, such a fix can be deferred beyond this PR.

intl.emu
```diff
@@ -11,6 +11,172 @@
...
+        1. Let _entry_ be the first element of _preferredUnits_.
+        1. Let _unit_ be _entry_.[[Unit]].
+        1. Repeat, while _preferredUnits_ has more than one element,
+          1. Let _threshold_ be _entry_.[[Threshold]].
+          1. Let _convertedValue_ be ? ConvertUnitValue(_sourceValue_, _sourceUnit_, _unit_).
+          1. If _convertedValue_ is not finite or abs(ℝ(_convertedValue_)) ≥ _threshold_, return _unit_.
+          1. Remove the first element of _preferredUnits_.
+          1. Set _entry_ to the first element of _preferredUnits_.
+          1. Set _unit_ to _entry_.[[Unit]].
+        1. Return _unit_.
```
gibson042: These steps can be greatly simplified.
```suggestion
        1. Let _unit_ be ~unset~.
        1. For each element _entry_ of _preferredUnits_, do
          1. Set _unit_ to _entry_.[[Unit]].
          1. Let _convertedValue_ be ? ConvertUnitValue(_sourceValue_, _sourceUnit_, _unit_).
          1. If _convertedValue_ is not finite or abs(ℝ(_convertedValue_)) ≥ _entry_.[[Threshold]], return _unit_.
        1. Assert: _unit_ is a String.
        1. Return _unit_.
```

intl.emu
```diff
@@ -11,6 +11,172 @@
...
+      </emu-note>
+      <emu-note type="editor">
+        <p>
+          TODO: The UTS #35 algorithm includes fallbacking for regions based on
+          <a href="https://www.unicode.org/cldr/charts/latest/supplemental/territory_containment_un_m_49.html">UN M.49 territory containment</a>,
+          but the unit preference data does not (yet) include any other numeric regions than 001.
+          We'll probably need to implement that in order to be compatible with possible future CLDR updates.
+        </p>
+      </emu-note>
+    </emu-clause>
```
gibson042, line 148: Just looking ahead, I think we want a single operation to correspond with [Compute the preferred output unit](https://unicode.org/reports/tr35/tr35-info.html#compute-the-preferred-output-unit).
```suggestion
    <emu-clause id="sec-amount-getpreferredunits" type="abstract operation">
      <h1>
        GetPreferredUnits (
          _category_: a String,
          _usage_: a String,
          _region_: a String,
        ): either a normal completion containing a List of Records or a throw completion
      </h1>
      <dl class="header">
        <dt>description</dt>
        <dd>It determines the preferred units to use for the given category, usage, and region according to the <a href="https://unicode.org/reports/tr35/tr35-info.html#compute-the-preferred-output-unit">Compute the preferred output unit</a> algorithm specified in <a href="https://unicode.org/reports/tr35/tr35-info.html#Unit_Preferences">Unicode Technical Standard #35 Part 6 Supplemental, Unit Preferences</a>, and returns them ordered by descending threshold.</dd>
      </dl>
      <emu-alg>
        1. Assert: _region_ can be matched by the <code>unicode_region_id</code> Unicode locale nonterminal and the ASCII-uppercase of _region_ is _region_.
        1. Let _candidateElements_ be ~unset~.
        1. Repeat, while _candidateElements_ is ~unset~,
          1. If _category_ is the <code>category</code> and _usage_ is the <code>usage</code> of a <code>&lt;unitPreferences&gt;</code> element in the <emu-xref href="#sec-amount-unit-preference-data">unit preference data</emu-xref>, then
            1. Let _unitPreferencesElement_ be that element.
            1. Set _candidateElements_ to the List of <code>&lt;unitPreference&gt;</code> elements in _unitPreferencesElement_.
          1. Else,
            1. Assert: _usage_ is not *"default"*.
            1. Let _len_ be the length of _usage_.
            1. Let _idx_ be StringLastIndexOf(_usage_, *"-"*, _len_ - 1).
            1. If _idx_ is ~not-found~ or _idx_ = 0, set _usage_ to *"default"*; else set _usage_ to the substring of _usage_ from 0 to _idx_.
        1. Let _matchingUnits_ be a new empty List.
        1. Repeat, while _matchingElements_ is empty,
          1. For each element _unitPreference_ of _candidateElements_, do
            1. Assert: _unitPreference_ has an attribute <code>regions</code>.
            1. Let _regionsAttr_ be the String value of the <code>regions</code> attribute of _unitPreference_.
            1. Let _regions_ be StringSplitToList(_regionsAttr_, *" "*).
            1. Let _geq_ be the mathematical value of the <code>geq</code> attribute of _unitPreference_ if present, or 1 otherwise.
            1. Let _unit_ be the String value of the body of the _unitPreference_.
            1. If _regions_ contains _region_, append the Record { [[Unit]]: _unit_, [[Threshold]]: _geq_ } to _matchingUnits_.
          1. If _matchingElements_ is empty, then
            1. Assert: _region_ is not *"001"*.
            1. Set _region_ to *"001".
        1. Return _matchingUnits_.
      </emu-alg>
      <emu-note>
        <p><a href="https://unicode.org/reports/tr35/tr35-info.html#constraints">Unicode Technical Standard #35 Part 1, Unit Preferences Constraints</a> specifies that the <emu-xref href="#sec-amount-unit-preference-data">unit preference data</emu-xref> for a given _category_, _usage_, and _region_ is non-empty and already in descending order [by threshold].</p>
      </emu-note>
      <emu-note type="editor">
        <p>
          TODO: The UTS #35 algorithm includes region fallback based on
          <a href="https://www.unicode.org/cldr/charts/latest/supplemental/territory_containment_un_m_49.html">UN M.49 territory containment</a>,
          but this algorithm currently only falls back to region <code>001</code>.
          The unit preference data does not (yet) include any other numeric regions, but we'll probably need revision in order to be compatible with possible future CLDR updates.
        </p>
      </emu-note>
    </emu-clause>
```

gibson042 (approved):
I'd like to name the Records that are passed between operations, but won't insist on it being part of this PR. The rest of this review is a large handful of non-blocking suggestions.

intl.emu
jessealama: Good idea. I've rearranged things.

spec.emu
jessealama: Good point. I've updated the description.

spec.emu
jessealama: Nice. I've added that.

spec.emu
jessealama: I've taken care of the observable order of operations here, consistent with #92 , but that issue remains open and deserves its own PR.

intl.emu
jessealama: I've made a follow-on issue (#151) to tackle this.

intl.emu
```diff
@@ -11,6 +11,143 @@
...
+
+    <emu-clause id="sec-amount-unit-preference-data">
+      <h1>Unit Preference Data</h1>
+      <p>Unit preference data is derived from CLDR file <a href="https://github.com/unicode-org/cldr/blob/main/common/supplemental/units.xml"><code>supplemental/units.xml</code></a>. As described in <a href="https://unicode.org/reports/tr35/tr35-info.html#Unit_Preferences_Data">Unicode Technical Standard #35 Part 6 Supplemental, Unit Preferences Data</a>, each <code>&lt;unitPreferences&gt;</code> element contains the preferred units for its <code>category</code> and <code>usage</code>. A <code>&lt;unitPreferences&gt;</code> element contains one or more <code>&lt;unitPreference&gt;</code> elements, which define the <code>regions</code> (a space-separated list of region identifiers) and optionally the <code>geq</code> threshold for the unit preference.</p>
+    </emu-clause>
+
+    <emu-clause id="sup-amount-selecttargetunit" type="abstract operation">
+      <h1>
+        SelectTargetUnit (
+          _sourceValue_: a Number,
```
gibson042: This input should probably be a mathematical value. https://github.com/unicode-org/cldr/blob/main/common/supplemental/units.xml contains `<unitPreference regions="001" geq="0.1">kilometer</unitPreference>`, and we're just lucky that the corresponding IEEE 754 binary64 value is slightly above 0.1 rather than slightly below it (otherwise it would incorrectly not be selected for **0.1**<sub>𝔽</sub>).
```suggestion
          _sourceValue_: an extended mathematical value or *NaN*,
```
(and likewise throughout this PR, including in spec.emu)

spec.emu
```diff
@@ -139,6 +139,53 @@ location: https://tc39.es/proposal-amount/
...
+        <dl class="header">
+          <dt>description</dt>
+          <dd>It validates the given _opts_ for translating an Amount's value to a new unit and returns their resolved values. A String _opts_ is treated as a convenient shorthand for specifying a unit.</dd>
+        </dl>
+        <emu-alg>
+          1. If _opts_ is a String, return ? GetAmountOptions(_opts_).
+          1. If _opts_ is not an Object, throw a *TypeError* exception.
+          1. Let _resolvedOpts_ be ? GetAmountOptions(_opts_).
+          1. Let _unitOption_ be _resolvedOpts_.[[Unit]].
+          1. Let _unit_ be ? SelectTargetUnit(_sourceValue_, _sourceUnit_, _opts_, _unitOption_).
```
gibson042: Nit:
```suggestion
          1. Let _unit_ be ? SelectTargetUnit(_sourceValue_, _sourceUnit_, _opts_, _resolvedOpts_.[[Unit]]).
```

spec.emu
```diff
@@ -459,9 +498,16 @@ location: https://tc39.es/proposal-amount/
...
         1. Let _sourceValue_ be StringToNumber(_v_).
+      1. Let _sourceUnit_ be _O_.[[Unit]].
+      1. If _sourceUnit_ is *undefined*, throw a *TypeError* exception.
+      1. Let _opts_ be ? GetAmountConvertToOptions(_sourceValue_, _sourceUnit_, _options_).
+      1. Let _targetUnit_ be _opts_.[[Unit]].
       1. Let _convertedValue_ be ? ConvertUnitValue(_sourceValue_, _sourceUnit_, _targetUnit_).
       1. Let _result_ be OrdinaryObjectCreate(%Amount.prototype%, « [[AmountValue]], [[Unit]] »).
+      1. Let _fractionDigits_ be _opts_.[[FractionDigits]].
+      1. Let _significantDigits_ be _opts_.[[SignificantDigits]].
       1. If _convertedValue_ is finite and (_fractionDigits_ is not *undefined* or _significantDigits_ is not *undefined*), then
```
gibson042: Noticed in passing:
```suggestion
      1. If _convertedValue_ is a finite Number and (_fractionDigits_ is not *undefined* or _significantDigits_ is not *undefined*), then
```

intl.emu
```diff
@@ -11,6 +11,143 @@
...
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It determines the appropriate unit identifier to use as the conversion target.</dd>
+        <dt>redefinition</dt>
+        <dd>true</dd>
+      </dl>
+      <p>This definition supersedes the definition provided in <emu-xref href="#sec-amount-selecttargetunit"></emu-xref>.</p>
+      <emu-alg>
+        1. Let _locale_ be ? Get(_options_, *"locale"*).
+        1. Let _usage_ be ? GetOption(_options_, *"usage"*, ~string~, ~empty~, *undefined*).
```
gibson042: This will result in observably reading "locale" and "usage" after "fractionDigits", "roundingMode", "significantDigits", and "unit" (i.e., out of alphabetical order) if and only if the implementation includes ECMA-402. See #95 more a more complete writeup.

intl.emu
```diff
@@ -11,6 +11,143 @@
...
+        1. If _unitOption_ is a String, then
+          1. If _usage_ is not *undefined* or _locale_ is not *undefined*, throw a *TypeError* exception.
+          1. Return _unitOption_.
+        1. If _usage_ is *undefined*, then
+          1. If _locale_ is *undefined*, throw a *TypeError* exception.
+          1. Set _usage_ to *"default"*.
+        1. Let _category_ be ? GetUnitCategory(_sourceUnit_).
+        1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locale_).
+        1. If _requestedLocales_ is an empty List, let _resolvedLocale_ be DefaultLocale(); else let _resolvedLocale_ be the first element of _requestedLocales_.
```
gibson042, line 97: This should use actual locale negotiation rather than just picking the first element; again, see #95.

intl.emu
```diff
@@ -11,6 +11,143 @@
...
+        1. Let _maximalLocale_ be the result of the <a href="https://unicode.org/reports/tr35/#Likely_Subtags">Add Likely Subtags</a> algorithm applied to _resolvedLocale_. If an error is signaled, set _maximalLocale_ to _resolvedLocale_.
+        1. Let _region_ be GetLocaleRegion(_maximalLocale_).
+        1. If _region_ is *undefined*, set _region_ to *"001"*.
+        1. Let _preferredUnits_ be GetPreferredUnits(_category_, _usage_, _region_).
+        1. Let _unit_ be ~unset~.
+        1. For each element _entry_ of _preferredUnits_, do
+          1. Set _unit_ to _entry_.[[Unit]].
+          1. Let _convertedValue_ be ? ConvertUnitValue(_sourceValue_, _sourceUnit_, _unit_).
```
gibson042, line 104: It's a bit awkward to specify O(n) ConvertUnitValue calls (n - 1 in SelectTargetUnit and the last in `Amount.prototype.convertTo`). I'm not objecting, but we should look for ways to improve that.

intl.emu
```diff
@@ -11,6 +11,143 @@
...
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It determines the preferred units to use for the given category, usage, and region according to the <a href="https://unicode.org/reports/tr35/tr35-info.html#compute-the-preferred-output-unit">Compute the preferred output unit</a> algorithm specified in <a href="https://unicode.org/reports/tr35/tr35-info.html#Unit_Preferences">Unicode Technical Standard #35 Part 6 Supplemental, Unit Preferences</a>, and returns them ordered by descending threshold.</dd>
+      </dl>
+      <emu-alg>
+        1. Assert: _region_ can be matched by the <code>unicode_region_id</code> Unicode locale nonterminal and the ASCII-uppercase of _region_ is _region_.
+        1. Let _candidateElements_ be ~unset~.
+        1. Repeat, while _candidateElements_ is ~unset~,
+          1. If _category_ is the <code>category</code> and _usage_ is the <code>usage</code> of a <code>&lt;unitPreferences&gt;</code> element in the <emu-xref href="#sec-amount-unit-preference-data">unit preference data</emu-xref>, then
+            1. Let _unitPreferencesElement_ be that element.
```
gibson042: Wording suggestion:
```suggestion
          1. [declared="unitPreferencesElement"] If there is a <code>&lt;unitPreferences&gt;</code> element _unitPreferencesElement_ in the <emu-xref href="#sec-amount-unit-preference-data">unit preference data</emu-xref> such that its <code>category</code> is _category_ and its <code>usage</code> is _usage_, then
```

intl.emu
```diff
@@ -11,6 +11,143 @@
...
+            1. Assert: _unitPreference_ has an attribute <code>regions</code>.
+            1. Let _regionsAttr_ be the String value of the <code>regions</code> attribute of _unitPreference_.
+            1. Let _regions_ be StringSplitToList(_regionsAttr_, *" "*).
+            1. Let _geq_ be the mathematical value of the <code>geq</code> attribute of _unitPreference_ if present, or 1 otherwise.
+            1. Let _unit_ be the String value of the body of the _unitPreference_.
+            1. If _regions_ contains _region_, append the Record { [[Unit]]: _unit_, [[Threshold]]: _geq_ } to _matchingUnits_.
+          1. If _matchingUnits_ is empty, then
+            1. Assert: _region_ is not *"001"*.
+            1. Set _region_ to *"001"*.
+        1. Return _matchingUnits_.
```
gibson042, line 171: https://unicode.org/reports/tr35/tr35-info.html#Unit_Preferences_Data:~:text=the%20last%20unitPreference%20is%20effectively%20geq%3D%220%22
> the last unitPreference is effectively geq="0"
```suggestion
        1. Let _lastUnit_ be the last element of _matchingUnits_.
        1. Set _lastUnit_.[[Threshold]] to 0.
        1. Return _matchingUnits_.
```

intl.emu
```diff
@@ -202,7 +341,7 @@
           1. Set _intlObj_.[[CurrencyDisplay]] to _currencyDisplay_.
           1. Set _intlObj_.[[CurrencySign]] to _currencySign_.
         1. If _style_ is *"unit"*, then
-          1. <ins>If _unit_ is not *undefined*, </ins> <del>Set</del> <ins>set</ins> _intlObj_.[[Unit]] to _unit_.
+          1. <ins>If _unit_ is not *undefined*,</ins> <del>Set</del> <ins>set</ins> _intlObj_.[[Unit]] to _unit_.
```
gibson042: Suggestion: Minimize the count of \<ins>/\<del> elements by using longer spans.
```suggestion
          1. <del>Set</del> <ins>If _unit_ is not *undefined*, set</ins> _intlObj_.[[Unit]] to _unit_.
```

gibson042 (changes requested):
I think this is still slightly too far off from #95.

intl.emu
jessealama: Ah, good catch, thanks.

intl.emu
jessealama: I've taken a look at this and added a couple of new AOs and a couple of additional fields to the definition of Amount Options Record. So I think the out-of-order issue should be resolved. Care to take another look?

intl.emu
jessealama: Agree. I propose to punt that to a separate branch that tackles #95.

intl.emu
jessealama: Agree. Would you like to file a follow-on issue to track this?

intl.emu
jessealama: I've made that change; thanks. (Note that we also have #125, which is still a draft, for tackling the bulk of the unit conversion precision issues at play here. But as for this change toward possibly-NaN extended MVs, it's done.)

spec.emu
```diff
@@ -103,27 +103,49 @@ location: https://tc39.es/proposal-amount/
...
+
       <emu-clause id="sec-amount-getamountoptions" type="abstract operation">
         <h1>GetAmountOptions (
-          _opts_: an ECMAScript language value
-        ): either a normal completion containing a Record with fields [[FractionDigits]] (a non-negative integer or *undefined*), [[RoundingMode]] (a rounding mode), [[SignificantDigits]] (a positive integer or *undefined*), and [[Unit]] (a String or *undefined*) or a throw completion
+          _opts_: an ECMAScript language value,
+          optional _sourceValue_: an extended mathematical value or *NaN*,
+          optional _sourceUnit_: a String,
+        ): either a normal completion containing an Amount Options Record or a throw completion
         </h1>
```
eemeli, line 125: This ends up feeling unnecesarily complex now, unless there's a real reason to guarantee that the option reading and processing order is fixed and alphabetic? What I'm thinking of here is separating this so that this AO only handles the parts that are common to construction and conversion, and there's a separate AO that's only called during conversion, which then reads the _locale_ and _usage_ from _opts_.

This might be easier for a spec reader to understand, and would provide a singular jumping-off point for the behaviour that's overridden by 402, rather than the three that are now interspersed here.

This would also allow for [[Locale]] and [[Usage]] to be left out of the Amount Options Record, as the only place where they're read is in the 402 version of SelectTargetUnit.

I've not read through all @gibson042's review comments, and he may well have recommended against what I'm describing here.

spec.emu
```diff
@@ -443,25 +538,31 @@ location: https://tc39.es/proposal-amount/
...
-      1. Let _significantDigits_ be _validatedOpts_.[[SignificantDigits]].
       1. Let _v_ be _O_.[[AmountValue]].
       1. If _v_ is a Number, then
-        1. Let _sourceValue_ be _v_.
+        1. Let _numericValue_ be _v_.
+        1. If _v_ is *NaN*, let _sourceValue_ be *NaN*; else let _sourceValue_ be the extended mathematical value of _v_.
       1. Else if _v_ is a BigInt, then
-        1. Let _sourceValue_ be 𝔽(ℝ(_v_)).
+        1. Let _numericValue_ be 𝔽(ℝ(_v_)).
+        1. Let _sourceValue_ be ℝ(_v_).
```
eemeli:
```suggestion
        1. Let _sourceValue_ be ℝ(_v_).
        1. Let _numericValue_ be 𝔽(_sourceValue_).
```

spec.emu
```diff
@@ -443,25 +538,31 @@ location: https://tc39.es/proposal-amount/
...
       1. Else,
         1. Assert: _v_ is a String.
-        1. Let _sourceValue_ be StringToNumber(_v_).
-      1. Let _convertedValue_ be ? ConvertUnitValue(_sourceValue_, _sourceUnit_, _targetUnit_).
+        1. Let _numericValue_ be StringToNumber(_v_).
+        1. Let _parsed_ be ParseText(_v_, |StringNumericLiteral|).
+        1. Assert: _parsed_ is a Parse Node.
+        1. Let _sourceMV_ be the StringIntlMV of _parsed_.
+        1. If _sourceMV_.[[Value]] is ~negative-zero~, let _sourceValue_ be 0; else let _sourceValue_ be _sourceMV_.[[Value]].
```
eemeli: Is there a reason to parse the string twice? Could we not determine _numericValue_ from _sourceValue_?
```suggestion
        1. Let _parsed_ be ParseText(_v_, |StringNumericLiteral|).
        1. Assert: _parsed_ is a Parse Node.
        1. Let _sourceMV_ be the StringIntlMV of _parsed_.
        1. If _sourceMV_.[[Value]] is ~negative-zero~, let _sourceValue_ be 0; else let _sourceValue_ be _sourceMV_.[[Value]].
        1. Let _numericValue_ be 𝔽(_sourceValue_).
```

spec.emu
```diff
@@ -443,25 +538,31 @@ location: https://tc39.es/proposal-amount/
...
+      1. NOTE: _sourceValue_ is the exact value of this Amount and is used to select the conversion target. _numericValue_ is its Number approximation and is used for the conversion itself.
```
eemeli: Given my comments above, couldn't _numericValue_ be determined at this point from _sourceValue_?
```suggestion
      1. Let _numericValue_ be 𝔽(_sourceValue_).
      1. NOTE: _sourceValue_ is the exact value of this Amount and is used to select the conversion target. _numericValue_ is its Number approximation and is used for the conversion itself.
```

intl.emu
eemeli: Nope, for conversion every entry includes a `001` region, which means that the region of a valid locale will _always_ be supported, and that therefore we will never fall back to a second-choice locale.

This does raise the question of whether we should have a `region` option instead of a `locale` option, as the data is region-dependent rather than language-dependent.

spec.emu
jessealama: Agree that this isn't particularly great. The only motivation is to ensure an alphabetical order of option reading. We can do some juggling to reduce the number of places that are clearly just 402 option-ordering artifacts. I'm happy to go back to the drawing board. Another alternative would be to tackle the issue in #95 .

jessealama: @gibson042 How far do we want to go in ensuring that a 402-compliant implementation reads options in alphabetical order? What I'm doing here is based perhaps on my overinterpretation of "read options in alphabetical order". We are writing 262 spec text to be agnostic about consulting the `usage` and `locale` options, which are specific to 402. But doing so requires introducing a layer of indirection in 262.

Another way to look at the option-ordering concern is to understand that 402 _adds_ new functionality (specifically, looking for and processing options left unspecified in 262), possibly with its own conventions enforced in its own realm, so to speak. This is somewhat more liberal than insisting that the 262 part be written defensively, anticipating (but not specifying) a 402-compliant implementation, working hard to ensure that a 402-compliant implementation can be slotted in and its conventions respected, even if 262 does not share those conventions.

gibson042: > Another way to look at the option-ordering concern is to understand that 402 _adds_ new functionality (specifically, looking for and processing options left unspecified in 262), possibly with its own conventions enforced in its own realm, so to speak.

This is not the model of either specification, nor is it the model of implementations (or even more importantly, of users). From a specification perspective, behavior must make sense both in ECMA-262 and in ECMA-402, because any given implementation is either going to cover just the former or both together.

But I agree that this approach feels overly complex. What about moving all observable reads into a single SetAmountOptions operation to be redefined by ECMA-402, and reserving a single Amount Options Record field for it rather than presuming it will need exactly [[Locale]] and [[Usage]]?
```
        <h1>GetAmountOptions (
          _opts_: an ECMAScript language value,
          optional _conversionSource_: an Amount,
        ): either a normal completion containing an Amount Options Record or a throw completion
        </h1>
        <dl class="header">
          <dt>description</dt>
          <dd>It validates the given _opts_ and returns their resolved values. A String _opts_ is treated as a shorthand for specifying only a unit. If _conversionSource_ is present, it can be used to select an otherwise unspecified target unit.</dd>
        </dl>
        <emu-alg>
          1. Assert: If _conversionSource_ is present, then _conversionSource_.[[Unit]] is not *undefined*.
          1. Let _resolvedOpts_ be the Amount Options Record { [[FractionDigits]]: *undefined*, [[RoundingMode]]: *"halfEven"*, [[SignificantDigits]]: *undefined*, [[Unit]]: *undefined*, [[ExtendedOptions]]: *undefined* }.
          1. If _opts_ is a String, then
            1. If ParseText(_opts_, |UnitIdentifier|) is not a Parse Node, throw a *RangeError* exception.
            1. Set _resolvedOpts_.[[Unit]] to _opts_.
            1. Return _resolvedOpts_.
          1. Perform ? SetAmountOptions(_resolvedOpts_, ? GetOptionsObject(_opts_)).
          1. Let _fractionDigits_ be _resolvedOpts_.[[FractionDigits]].
          1. Let _significantDigits_ be _resolvedOpts_.[[SignificantDigits]].
          1. If _fractionDigits_ is not *undefined*, then
            …
          1. Else if _significantDigits_ is not *undefined*, then
            …
          1. If _conversionSource_ is present, set _resolvedOpts_.[[Unit]] to ? SelectTargetUnit(_conversionSource_, _resolvedOpts_).
          1. Return _resolvedOpts_.
        </emu-alg>
```
```
        <h1>
          SetAmountOptions (
            _target_: an Amount Options Record,
            _source_: an Object,
          ): either a normal completion containing ~unused~ or a throw completion
        </h1>
        <dl class="header">
          <dt>description</dt>
          <dd>It sets the fields of _target_ from _source_, reading properties in alphabetical order.</dd>
        </dl>
        <p>
          An ECMAScript implementation that includes the ECMA-402 Internationalization API
          must implement this method as specified in the ECMA-402 specification.
          If an ECMAScript implementation does not include the ECMA-402 API,
          SetAmountOptions performs the following steps when called:
        </p>
        <emu-alg>
            1. Set _target_.[[FractionDigits]] to ? GetOption(_source_, *"fractionDigits"*, ~number~, ~empty~, *undefined*).
            1. Set _target_.[[RoundingMode]] to ? GetOption(_source_, *"roundingMode"*, ~string~, « *"ceil"*, *"floor"*, *"expand"*, *"trunc"*, *"halfCeil"*, *"halfFloor"*, *"halfExpand"*, *"halfTrunc"*, *"halfEven"* », *"halfEven"*).
            1. Set _target_.[[SignificantDigits]] to ? GetOption(_source_, *"significantDigits"*, ~number~, ~empty~, *undefined*).
            1. Let _unit_ be ? GetOption(_source_, *"unit"*, ~string~, ~empty~, *undefined*).
            1. If _unit_ is not *undefined*, then
              1. If ParseText(_opts_, |UnitIdentifier|) is not a Parse Node, throw a *RangeError* exception.
            1. Set _target_.[[Unit]] to _unit_.
            1. Return ~unused~.
        </emu-alg>
```
(the above also gratuitously performs immediate validation of unit syntax and simplifies the signature of GetAmountOptions and SelectTargetUnit)

spec.emu
```diff
@@ -443,25 +538,32 @@ location: https://tc39.es/proposal-amount/
...
+        1. Let _numericValue_ be 𝔽(_sourceValue_).
       1. Else,
         1. Assert: _v_ is a String.
-        1. Let _sourceValue_ be StringToNumber(_v_).
-      1. Let _convertedValue_ be ? ConvertUnitValue(_sourceValue_, _sourceUnit_, _targetUnit_).
+        1. Let _parsed_ be ParseText(_v_, |StringNumericLiteral|).
+        1. Assert: _parsed_ is a Parse Node.
+        1. Let _sourceMV_ be the StringIntlMV of _parsed_.
+        1. If _sourceMV_.[[Value]] is ~negative-zero~, let _sourceValue_ be 0; else let _sourceValue_ be _sourceMV_.[[Value]].
+        1. Let _numericValue_ be 𝔽(_sourceValue_).
```
gibson042: Isn't this lossy for a value string like "-0"? IEEE 754 binary64 multiplication preserves the sign of a zero, so we want e.g. `new Amount("-0", "meter").convertTo("foot")` to remain negative.

spec.emu
```diff
@@ -443,25 +538,32 @@ location: https://tc39.es/proposal-amount/
...
+      1. Let _numericValue_ be 𝔽(_sourceValue_).
```
gibson042: _numericValue_ is initialized in each branch above, and then unconditionally overwritten here. Please pick one.

spec.emu
```diff
@@ -443,25 +538,32 @@ location: https://tc39.es/proposal-amount/
...
+      1. Let _sourceUnit_ be _O_.[[Unit]].
+      1. If _sourceUnit_ is *undefined*, throw a *TypeError* exception.
+      1. Let _opts_ be ? GetAmountOptions(_options_, _sourceValue_, _sourceUnit_).
+      1. Let _targetUnit_ be _opts_.[[Unit]].
+      1. Let _convertedValue_ be ? ConvertUnitValue(_numericValue_, _sourceUnit_, _targetUnit_).
       1. Let _result_ be OrdinaryObjectCreate(%Amount.prototype%, « [[AmountValue]], [[Unit]] »).
-      1. If _convertedValue_ is finite and (_fractionDigits_ is not *undefined* or _significantDigits_ is not *undefined*), then
+      1. Let _fractionDigits_ be _opts_.[[FractionDigits]].
+      1. Let _significantDigits_ be _opts_.[[SignificantDigits]].
+      1. If _convertedValue_ is a finite Number and (_fractionDigits_ is not *undefined* or _significantDigits_ is not *undefined*), then
```
gibson042: Suggestion in passing:
```suggestion
      1. If _fractionDigits_ is not *undefined* or _significantDigits_ is not *undefined*, let _needsRounding_ be *true*, else let _needsRounding_ be *false*.
      1. If _convertedValue_ is a finite Number and _needsRounding_ is *true*, then
```

spec.emu
jessealama: Good catch. I've fixed that.

spec.emu
jessealama: Fixed.

spec.emu
jessealama: That looks good to me. Thanks for clarifying. I've incorporated these changes. Care to give it another look?

intl.emu
jessealama: I've tweaked things so that we use `RegionPreference` here, which also picks up region. In light of the fact taht every category and usage has a `001` entry, no requested locale can go unsupported, so I'm not sure I see how negotiation should go. I may be missing something. Perhaps file an issue to tackle this?

On region versus locale, I'd lean toward keeping locale. UTS 35 takes a locale as input, and it matches how `Intl.Locale` handles region-based data (calendars, hour cycles). That strikes me as sufficient, or do we need more here?

README.md
```diff
@@ -130,6 +130,7 @@ The object prototype would provide the following methods:
     so `amount.convertTo("milliliter")` is the same as `amount.convertTo({ unit: "milliliter" })`.
   * `locale` (String or Array of Strings or undefined):
     The locale for which the preferred unit of the corresponding category is determined.
+    Unit preferences depend only on the region of the locale, which is taken from the `rg` Unicode extension key (if present)
```
eemeli: This is misleading, as e.g. the region of `en-GB` is `GB`.

jessealama: Good point. I've fleshed out the algorithm a bit. I didn't mention `"001"` in the README as an ultimate fallback. I think that's OK for a README.

README.md
```diff
@@ -130,6 +130,7 @@ The object prototype would provide the following methods:
     so `amount.convertTo("milliliter")` is the same as `amount.convertTo({ unit: "milliliter" })`.
   * `locale` (String or Array of Strings or undefined):
     The locale for which the preferred unit of the corresponding category is determined.
+    Unit preferences depend only on the region of the locale, determined as in ECMA-402's [RegionPreference](https://tc39.es/ecma402/#sec-regionpreference): (1) the `rg` Unicode extension, if present ('US' for `en-GB-u-rg-uszzzz`), (2) the locale's region subtag (`GB` for `en-GB`), and finally (3) the most likely region for the locale (`US` for `en`).
```
gibson042:
```suggestion
    Unit preferences depend only on the region of the locale, determined as in ECMA-402's [RegionPreference](https://tc39.es/ecma402/#sec-regionpreference): (1) the `rg` Unicode extension if present ("US" for `en-GB-u-rg-uszzzz`), (2) the locale's region subtag if present ("CH" for `de-CH`), the most likely region for the locale if known ("BR" for `pt`), and finally "001" (the World) as an ultimate fallback.
```

spec.emu
```diff
@@ -103,39 +103,131 @@ location: https://tc39.es/proposal-amount/
         </emu-alg>
       </emu-clause>
 
+      <emu-clause id="sec-amount-options-record">
+        <h1>Amount Options Record</h1>
+        <p>
+          An <dfn variants="Amount Options Records">Amount Options Record</dfn> is a Record holding the resolved values of the options
+          accepted by the Amount constructor and by <emu-xref href="#sec-amount.prototype.convertto">Amount.prototype.convertTo</emu-xref>.
```
gibson042: Simplification:
```suggestion
          accepted by the Amount constructor and by <emu-xref href="#sec-amount.prototype.convertto" title></emu-xref>.
```

spec.emu
```diff
@@ -103,39 +103,131 @@ location: https://tc39.es/proposal-amount/
...
+          1. Let _resolvedOpts_ be the Amount Options Record { [[FractionDigits]]: *undefined*, [[RoundingMode]]: *"halfEven"*, [[SignificantDigits]]: *undefined*, [[Unit]]: *undefined*, [[ExtendedOptions]]: *undefined* }.
           1. If _opts_ is a String, then
-            1. Let _unit_ be _opts_.
-          1. Else,
-            1. Set _opts_ to ? GetOptionsObject(_opts_).
-            1. Set _fractionDigits_ to ? GetOption(_opts_, *"fractionDigits"*, ~number~, ~empty~, *undefined*).
-            1. Set _roundingMode_ to ? GetOption(_opts_, *"roundingMode"*, ~string~, « *"ceil"*, *"floor"*, *"expand"*, *"trunc"*, *"halfCeil"*, *"halfFloor"*, *"halfExpand"*, *"halfTrunc"*, *"halfEven"* », *"halfEven"*).
-            1. Set _significantDigits_ to ? GetOption(_opts_, *"significantDigits"*, ~number~, ~empty~, *undefined*).
-            1. Let _unit_ be ? GetOption(_opts_, *"unit"*, ~string~, ~empty~, *undefined*).
+            1. If ParseText(_opts_, |UnitIdentifier|) is not a Parse Node, throw a *RangeError* exception.
```
gibson042: This may change based on discussion in next week's plenary, but I at least am advocating for
```suggestion
            1. If ParseText(_opts_, |UnitIdentifier|) is not a Parse Node, throw a *SyntaxError* exception.
```

spec.emu
```diff
@@ -103,39 +103,131 @@ location: https://tc39.es/proposal-amount/
...
+        <dl class="header">
+          <dt>description</dt>
+          <dd>It sets the fields of _target_ from the properties of _source_, which are read in alphabetical order. Validation that depends on more than one option is left to the caller.</dd>
+        </dl>
+        <p>
+          An ECMAScript implementation that includes the ECMA-402 Internationalization API
+          must implement this abstract operation as specified in the ECMA-402 specification.
+          If an ECMAScript implementation does not include the ECMA-402 API,
+          SetAmountOptions performs the following steps when called:
+        </p>
```
gibson042: Nit:
```suggestion
        <p>
          An ECMAScript implementation that includes the ECMA-402 Internationalization API
          must implement this abstract operation as specified in ECMA-402.
          Otherwise, SetAmountOptions performs the following steps when called:
        </p>
```
(mirroring e.g. [AvailableNamedTimeZoneIdentifiers](https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-availablenamedtimezoneidentifiers))

spec.emu
```diff
@@ -103,39 +103,131 @@ location: https://tc39.es/proposal-amount/
...
+          1. Return _resolvedOpts_.
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-amount-setamountoptions" type="abstract operation">
+        <h1>
+          SetAmountOptions (
+            _target_: an Amount Options Record,
+            _source_: an Object,
+          ): either a normal completion containing ~unused~ or a throw completion
```
gibson042: I think we want to read "locale" and "usage" for conversion but not construction, which means that distinction should be expressed in a parameter.
```suggestion
          SetAmountOptions (
            _target_: an Amount Options Record,
            _source_: an Object,
            _purpose_: ~construction~ or ~unit-conversion~,
          ): either a normal completion containing ~unused~ or a throw completion
```

spec.emu
```diff
@@ -103,39 +103,131 @@ location: https://tc39.es/proposal-amount/
...
+        <dl class="header">
+          <dt>description</dt>
+          <dd>It determines the unit identifier to which _conversionSource_ is to be converted.</dd>
+        </dl>
+        <p>
+          An ECMAScript implementation that includes the ECMA-402 Internationalization API
+          must implement this abstract operation as specified in the ECMA-402 specification.
+          If an ECMAScript implementation does not include the ECMA-402 API,
+          SelectTargetUnit performs the following steps when called:
+        </p>
```
gibson042:
```suggestion
        <p>
          An ECMAScript implementation that includes the ECMA-402 Internationalization API
          must implement this abstract operation as specified in ECMA-402.
          Otherwise, SelectTargetUnit performs the following steps when called:
        </p>
```

intl.emu
```diff
@@ -11,6 +11,179 @@
...
+      <h1>Amount Options Record</h1>
+      <p>In an implementation that includes the ECMA-402 Internationalization API, the [[ExtendedOptions]] field of an <emu-xref href="#sec-amount-options-record">Amount Options Record</emu-xref> that has been populated by SetAmountOptions is a Record with fields [[Locale]] (an ECMAScript language value) and [[Usage]] (a String or *undefined*).</p>
+    </emu-clause>
+
+    <emu-clause id="sup-amount-setamountoptions" type="abstract operation">
+      <h1>
+        SetAmountOptions (
+          _target_: an Amount Options Record,
+          _source_: an Object,
+        ): either a normal completion containing ~unused~ or a throw completion
```
gibson042:
```suggestion
        SetAmountOptions (
          _target_: an Amount Options Record,
          _source_: an Object,
          _purpose_: ~construction~ or ~unit-conversion~,
        ): either a normal completion containing ~unused~ or a throw completion
```

intl.emu
```diff
@@ -11,6 +11,179 @@
...
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It sets the fields of _target_ from the properties of _source_, which are read in alphabetical order. Validation that depends on more than one option is left to the caller.</dd>
+        <dt>redefinition</dt>
+        <dd>true</dd>
+      </dl>
+      <p>This definition supersedes the definition provided in <emu-xref href="#sec-amount-setamountoptions"></emu-xref>.</p>
+      <emu-alg>
+        1. Set _target_.[[FractionDigits]] to ? GetOption(_source_, *"fractionDigits"*, ~number~, ~empty~, *undefined*).
+        1. Let _locale_ be ? Get(_source_, *"locale"*).
```
gibson042: In addition to only being relevant for unit conversion, `locale` should be validated immediately as it is by e.g. [ResolveOptions](https://tc39.es/ecma402/#sec-resolveoptions).
```suggestion
        1. If _purpose_ is ~unit-conversion~, then
          1. Let _localeValue_ be ? Get(_source_, *"locale"*).
          1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_localeValue_).
```

intl.emu
```diff
@@ -11,6 +11,179 @@
...
+        1. Set _target_.[[RoundingMode]] to ? GetOption(_source_, *"roundingMode"*, ~string~, « *"ceil"*, *"floor"*, *"expand"*, *"trunc"*, *"halfCeil"*, *"halfFloor"*, *"halfExpand"*, *"halfTrunc"*, *"halfEven"* », *"halfEven"*).
+        1. Set _target_.[[SignificantDigits]] to ? GetOption(_source_, *"significantDigits"*, ~number~, ~empty~, *undefined*).
+        1. Let _unit_ be ? GetOption(_source_, *"unit"*, ~string~, ~empty~, *undefined*).
+        1. If _unit_ is not *undefined*, then
+          1. If ParseText(_unit_, |UnitIdentifier|) is not a Parse Node, throw a *RangeError* exception.
+        1. Set _target_.[[Unit]] to _unit_.
+        1. Let _usage_ be ? GetOption(_source_, *"usage"*, ~string~, ~empty~, *undefined*).
```
gibson042: Likewise `usage`.
```suggestion
        1. If _purpose_ is ~unit-conversion~, let _usage_ be ? GetOption(_source_, *"usage"*, ~string~, ~empty~, *undefined*).
```

intl.emu
```diff
@@ -11,6 +11,179 @@
...
+        1. Set _target_.[[ExtendedOptions]] to the Record { [[Locale]]: _locale_, [[Usage]]: _usage_ }.
```
gibson042: And likewise `usage`.
```suggestion
        1. If _purpose_ is ~unit-conversion~, then
          1. Let _usage_ be ? GetOption(_source_, *"usage"*, ~string~, ~empty~, *undefined*).
          1. Set _target_.[[ExtendedOptions]] to the Record { [[Locales]]: _requestedLocales_, [[Usage]]: _usage_ }.
```

intl.emu
```diff
@@ -11,6 +11,179 @@
...
+    </emu-clause>
+
+    <emu-clause id="sup-amount-options-record">
+      <h1>Amount Options Record</h1>
+      <p>In an implementation that includes the ECMA-402 Internationalization API, the [[ExtendedOptions]] field of an <emu-xref href="#sec-amount-options-record">Amount Options Record</emu-xref> that has been populated by SetAmountOptions is a Record with fields [[Locale]] (an ECMAScript language value) and [[Usage]] (a String or *undefined*).</p>
```
gibson042:
```suggestion
      <p>In an implementation that includes the ECMA-402 Internationalization API, the [[ExtendedOptions]] field of an <emu-xref href="#sec-amount-options-record">Amount Options Record</emu-xref> that has been populated by SetAmountOptions is a Record with fields [[Locales]] (a Language Priority List or *undefined*) and [[Usage]] (a String or *undefined*).</p>
```

intl.emu
```diff
@@ -11,6 +11,179 @@
...
+          1. If _usage_ is not *undefined* or _locale_ is not *undefined*, throw a *TypeError* exception.
+          1. Return _unit_.
+        1. If _usage_ is *undefined*, then
+          1. If _locale_ is *undefined*, throw a *TypeError* exception.
+          1. Set _usage_ to *"default"*.
+        1. Let _sourceUnit_ be _conversionSource_.[[Unit]].
+        1. Let _sourceValue_ be GetAmountNumericValue(_conversionSource_).
+        1. Let _category_ be ? GetUnitCategory(_sourceUnit_).
+        1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locale_).
+        1. If _requestedLocales_ is an empty List, let _resolvedLocale_ be DefaultLocale(); else let _resolvedLocale_ be the first element of _requestedLocales_.
```
gibson042:
```suggestion
        1. Let _requestedLocales_ be _extendedOpts_.[[Locales]].
        1. Let _usage_ be _extendedOpts_.[[Usage]].
        1. If _unit_ is not *undefined*, then
          1. If _usage_ is not *undefined* or _requestedLocales_ is not *undefined*, throw a *TypeError* exception.
          1. Return _unit_.
        1. If _usage_ is *undefined*, then
          1. If _requestedLocales_ is *undefined*, throw a *TypeError* exception.
          1. Set _usage_ to *"default"*.
        1. Let _sourceUnit_ be _conversionSource_.[[Unit]].
        1. Let _sourceValue_ be GetAmountNumericValue(_conversionSource_).
        1. Let _category_ be ? GetUnitCategory(_sourceUnit_).
        1. If _requestedLocales_ is an empty List, let _resolvedLocale_ be DefaultLocale(); else let _resolvedLocale_ be the first element of _requestedLocales_.
```

spec.emu
```diff
@@ -103,39 +103,131 @@ location: https://tc39.es/proposal-amount/
...
+          1. Return _unit_.
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-amount-getamountnumericvalue" type="abstract operation">
+        <h1>
+          GetAmountNumericValue (
+            _amount_: an Amount,
+          ): a Number
+        </h1>
```
gibson042: Nit: ECMA-262 often uses "numeric" to encompass the union of Number and BigInt.
```suggestion
      <emu-clause id="sec-amount-getamountnumbervalue" type="abstract operation">
        <h1>
          GetAmountNumberValue (
            _amount_: an Amount,
          ): a Number
        </h1>
```

intl.emu
```diff
@@ -11,6 +11,179 @@
...
+        1. Let _regionPreference_ be RegionPreference(_resolvedLocale_).
+        1. If _regionPreference_.[[RegionOverride]] is not *undefined*, let _region_ be _regionPreference_.[[RegionOverride]]; else let _region_ be _regionPreference_.[[Region]].
```
gibson042:
```suggestion
        1. NOTE: An unknown locale is associated with [[Region]] *"001"* (the World).
        1. If _regionPreference_.[[RegionOverride]] is not *undefined*, let _region_ be _regionPreference_.[[RegionOverride]]; else let _region_ be _regionPreference_.[[Region]].
```

intl.emu
```diff
@@ -11,6 +11,179 @@
...
+          _usage_: a String,
+          _region_: a String,
+        ): a List of Records
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It determines the preferred units to use for the given category, usage, and region according to the <a href="https://unicode.org/reports/tr35/tr35-info.html#compute-the-preferred-output-unit">Compute the preferred output unit</a> algorithm specified in <a href="https://unicode.org/reports/tr35/tr35-info.html#Unit_Preferences">Unicode Technical Standard #35 Part 6 Supplemental, Unit Preferences</a>, and returns them ordered by descending threshold.</dd>
+      </dl>
+      <emu-alg>
+        1. Assert: _region_ can be matched by the <code>unicode_region_id</code> Unicode locale nonterminal and the ASCII-uppercase of _region_ is _region_.
```
gibson042: Is this a typo? Cf. https://unicode.org/reports/tr35/#unicode_region_subtag .
```suggestion
        1. Assert: _region_ can be matched by the <code>unicode_region_subtag</code> Unicode locale nonterminal and the ASCII-uppercase of _region_ is _region_.
```

gibson042 (approved):
LGTM after a large handful of tweaks.

intl.emu
gibson042: #168

Issue #108, opened 2026-05-15
May review


waldemarhorwat:
Here's my review so far. I like the direction and approach. A number of problems remain with rounding which will make users believe that Amount can't be used to get reliably rounded results. Also, parts of the spec contradict itself and the explainer, and I'm not sure what the intent there is.

* My main observation from last meeting still stands: whatever behavior we standardize in the first released version should be the behavior we're happy with long-term, including for things like converting 2m to person-height in the US or France. We shouldn't make users set a flag in the future to get the good result rather than the one we standardized first. Releasing this simultaneously with Intl-Sequence-Units is one possible solution.
* The explainer and spec contradict each other in the `convertTo` API. The explainer has `unit` and `usage` options, the spec does not.
* The spec's note in the `Amount` constructor is incorrect: The spec always stores Number and BigInt arguments directly in [[AmountValue]], regardless of the rounding options.
* Exponents of 0 are not preserved when handing a string such as "0e-4". I'm happy with that behavior (and wouldn't want to change it) but wanted to make sure that's clear to everyone.
* Rounding continues to produce surprises:
  * Converting 5 grams to tonnes results in 0.0000049999999999999996 tonnes instead of 0.000005 tonnes. This is caused by double rounding when doing the conversion. There are techniques to do the conversion with just a single rounding at the end which would produce the exactly rounded result as long as the conversion factors aren't too large (i.e. we're not converting solar-masses to daltons).
  * Converting "0.15" meters to meters with conversion options fractionDigits=1 and the default rounding mode "halfEven" produces "1e-1" meters, which is incorrect. The correct answer is "2e-1" meters. Again, multiple rounding is the culprit here.

eemeli:
> The explainer and spec contradict each other in the `convertTo` API. The explainer has `unit` and `usage` options, the spec does not.

Presuming that you mean `locale` and `usage`, you're right; I filed #109 for this.

> Converting 5 grams to tonnes results in 0.0000049999999999999996 tonnes instead of 0.000005 tonnes. This is caused by double rounding when doing the conversion. There are techniques to do the conversion with just a single rounding at the end which would produce the exactly rounded result as long as the conversion factors aren't too large (i.e. we're not converting solar-masses to daltons).

Could you clarify what techniques you're referring to here? And would you agree that iterating on this would be appropriate to do within Stage 2?

I find it interesting that these operations don't produce the same results:
```js
5 * 1e-6 === 0.0000049999999999999996
[...]

PR #112, opened 2026-05-19
Be explicit about the Number-to-MV conversion in `.convertTo()`

eemeli:
Rather than getting an MV using ℝ(_convertedValue_), let's do the same thing Intl.NumberFormat does, and call ToIntlMathematicalValue(_convertedValue_), which effectively gets us ℝ(Number::toString(_convertedValue_, 10)).

This fixes the following part of #108:
> Converting "0.15" meters to meters with conversion options fractionDigits=1 and the default rounding mode "halfEven" produces "1e-1" meters, which is incorrect. The correct answer is "2e-1" meters. Again, multiple rounding is the culprit here.

### DIFF
--- a/spec.emu
+++ b/spec.emu
@@ -457,7 +457,8 @@ location: https://github.com/tc39/proposal-amount/
       1. Let _result_ be OrdinaryObjectCreate(%Amount.prototype%, « [[AmountValue]], [[Unit]] »).
       1. If _convertedValue_ is finite and (_fractionDigits_ is not *undefined* or _significantDigits_ is not *undefined*), then
         1. Let _formatter_ be CreateFormatterObject(_roundingMode_, _fractionDigits_, _fractionDigits_, _significantDigits_, _significantDigits_, *undefined*).
-        1. Let _formatted_ be FormatNumericToString(_formatter_, ℝ(_convertedValue_), 0).
+        1. Let _intlMV_ be ! ToIntlMathematicalValue(_convertedValue_).
+        1. Let _formatted_ be FormatNumericToString(_formatter_, _intlMV_.[[Value]], 0).
         1. Let _formattedMV_ be ! ToIntlMathematicalValue(_formatted_.[[FormattedString]]).
         1. Set _result_.[[AmountValue]] to RenderInExponentialNotation(_formattedMV_.[[Value]], _formattedMV_.[[StringDigitCount]]).
       1. Else,


sffc (approved):
A little surprising that `ToIntlMathematicalValue` is called twice, but I understand why.

jessealama (approved):

spec.emu
```diff
@@ -457,7 +457,8 @@ location: https://github.com/tc39/proposal-amount/
       1. Let _result_ be OrdinaryObjectCreate(%Amount.prototype%, « [[AmountValue]], [[Unit]] »).
       1. If _convertedValue_ is finite and (_fractionDigits_ is not *undefined* or _significantDigits_ is not *undefined*), then
         1. Let _formatter_ be CreateFormatterObject(_roundingMode_, _fractionDigits_, _fractionDigits_, _significantDigits_, _significantDigits_, *undefined*).
-        1. Let _formatted_ be FormatNumericToString(_formatter_, ℝ(_convertedValue_), 0).
+        1. Let _intlMV_ be ! ToIntlMathematicalValue(_convertedValue_).
+        1. Let _formatted_ be FormatNumericToString(_formatter_, _intlMV_.[[Value]], 0).
         1. Let _formattedMV_ be ! ToIntlMathematicalValue(_formatted_.[[FormattedString]]).
         1. Set _result_.[[AmountValue]] to RenderInExponentialNotation(_formattedMV_.[[Value]], _formattedMV_.[[StringDigitCount]]).
```
gibson042, line 463: I see four uses of RenderInExponentialNotation:
* in the [Amount constructor](https://tc39.es/proposal-amount/#sec-the-amount-constructor-value), with the fields of an Intl MV Record derived from a FormatNumericToString output string
* in [Amount.prototype.convertTo](https://tc39.es/proposal-amount/#sec-amount.prototype.convertto) (here), with the fields of an Intl MV Record derived from a FormatNumericToString output string
* in [Amount.prototype.toString](https://tc39.es/proposal-amount/#sec-amount.prototype.tostring), with the fields of an Intl MV Record derived from a Number
* in [Amount.prototype.toString](https://tc39.es/proposal-amount/#sec-amount.prototype.tostring), with the fields of an Intl MV Record derived from a BigInt::toString(_v_, 10) decimal digit sequence (_where those fields could alternatively be defined directly by counting the digits in the decimal representation of `_v_`, e.g. `ceiling(log10(abs(_v_) + 1))` or a named equivalent_)

The uses of FormatNumericToString seem to be necessary, but it occurs to me that we could probably remove a **lot** of translation, e.g. carrying non-exponential decimal strings from FormatNumericToString and maybe-exponential decimal strings from Number::toString all the way to the exponent-ensuring and decimal-point-shifting work of RenderInExponentialNotation without passing through non-string intermediaries. Let's save it for a followup, though.

Issue #115, opened 2026-05-21
Computing Correct Results


waldemarhorwat:
In this issue I'll explore the diverse sources of the rounding problems in unit conversions as well as how to fix them permanently. It's a bit of a long post, so I'll focus on the conversions that don't have an offset. There exist only two units that have offsets in the CLDR table (Celsius and Fahrenheit), and handling them won't be too difficult once we settle on the correct approach for units without an offset.

I hope that at some future time the language will include some kind of a Decimal type. When it does, Amount should be compatible with it — it would be bad if adding Decimal in the future would cause breaking changes to Amount. To ensure that, I'll sketch an Amount design that includes Decimal inputs (in addition to Number and BigInt) and we can decide which parts we can omit because we don't want to implement them today.

For clarity:
* I will suffix Numbers by 𝔽 and Decimals by 𝔻. Real numbers will have no suffix.
* I'm skipping the conversion to exponential notation for Strings.

## State Today

Let's look at a few conversions in today's implementation using Number inputs and outputs. Let's start with integral inputs that are representable exactly as Numbers:

| Input | Unit In | Unit Out | Conv Factor | Output |
| ----: | ----: |  ----: | ----: | ----: |
| 3<sub>𝔽</sub> | m | mm | 1000 | 3000<sub>𝔽</sub> |
| 5<sub>𝔽</sub> | in | cm | 127/50 | 12.7<sub>𝔽</sub> |
| 84<sub>𝔽</sub> | in | ft | 1/12 | 7<sub>𝔽</sub> |
| 5<sub>𝔽</sub> | g | tonnes | 1/1000000 | 0.0000049999999999999996<sub>𝔽</sub> |
| 825<sub>𝔽</sub> | g | kg | 1/1000 | 0.8250000000000001<sub>𝔽</sub> |

The incorrect outputs above are due to multiple rounding. As I'll show later, doing the conversion more precisely will fix these incorrect results.

Next let's take a look at some non-integral Number inputs:

| Input | MV(Input) | Unit In | Unit Out | Conv Factor | Output |
| ----: | :---- | ----: |  ----: | ----: | ----: |
| 0.003<sub>𝔽</sub> | 0.0030000000 0000000006 2450045135 1650553988 2928133010 8642578125 | m | mm | 1000 | 3<sub>𝔽</sub> |
[...]

PR #125, opened 2026-06-24
Use Intl mathematical values in `ConvertUnitValue`

jessealama:
Related to #115 , use (Intl) mathematical values in `ConvertUnitValue` to remove at least one step of rounding. Before:

```
Return _value_ × 𝔽(_sourceFactor_ / _targetFactor_)
```

Now:

```
Return 𝔽(ℝ(_value_) × _sourceFactor_ / _targetFactor_):
```


spec.emu
```diff
@@ -268,32 +268,35 @@ location: https://tc39.es/proposal-amount/
...
+          1. If _value_ is ~positive-infinity~ or _value_ is ~negative-infinity~, return _value_.
           1. Let _sourceFactor_ be _sourceConv_.[[Factor]].
           1. Let _sourceOffset_ be _sourceConv_.[[Offset]].
           1. Let _targetFactor_ be _targetConv_.[[Factor]].
           1. Let _targetOffset_ be _targetConv_.[[Offset]].
           1. If _sourceOffset_ is _targetOffset_, then
-            1. NOTE: This preserves a _value_ of *-0*<sub>𝔽</sub>.
-            1. Return _value_ × 𝔽(_sourceFactor_ / _targetFactor_).
-          1. Return _value_ × 𝔽(_sourceFactor_ / _targetFactor_) + 𝔽((_sourceOffset_ - _targetOffset_) / _targetFactor_).
+            1. NOTE: With equal offsets the conversion is purely multiplicative. Since unit factors are positive, this preserves a _value_ of ~negative-zero~.
```
gibson042, line 292: Nothing in https://unicode.org/reports/tr35/tr35-info.html#conversion-data or https://github.com/unicode-org/cldr/blob/main/common/supplemental/units.xml suggests to me that `factor` **MUST** be positive (and in fact I can easily imagine dimensions like angle or voltage having a unit with negative factor), so if this algorithm is going to make that assumption then it should be enforced with an Assert step.

spec.emu
```diff
@@ -268,32 +268,35 @@ location: https://tc39.es/proposal-amount/
...
-        ): either a normal completion containing a Number or a throw completion
+        ): either a normal completion containing either a mathematical value or one of ~positive-infinity~, ~negative-infinity~, ~not-a-number~, or ~negative-zero~, or a throw completion
         </h1>
         <dl class="header">
           <dt>description</dt>
-          <dd>It converts _value_ from _sourceUnit_ to _targetUnit_ using Number arithmetic.</dd>
+          <dd>It converts _value_ from _sourceUnit_ to _targetUnit_, returning the exact mathematical value of the result.</dd>
         </dl>
         <emu-alg>
           1. If SameValue(_sourceUnit_, _targetUnit_) is *true*, return _value_.
```
gibson042, line 281: I think this would be more readable if the input and output types were Intl MV Records.
```suggestion
          _intlMV_: an Intl MV Record,
          _sourceUnit_: a String,
          _targetUnit_: a String
        ): either a normal completion containing an Intl MV Record, or a throw completion
        </h1>
        <dl class="header">
          <dt>description</dt>
          <dd>It converts _value_ from _sourceUnit_ to _targetUnit_, returning the exact mathematical value of the result.</dd>
        </dl>
        <emu-alg>
          1. If SameValue(_sourceUnit_, _targetUnit_) is *true*, return _intlMV_.
```

spec.emu
```diff
@@ -268,32 +268,35 @@ location: https://tc39.es/proposal-amount/
...
           1. Let _sourceConv_ be ? GetUnitConversionFactor(_sourceUnit_).
           1. Let _targetConv_ be ? GetUnitConversionFactor(_targetUnit_).
           1. If _sourceConv_.[[BaseUnit]] is not _targetConv_.[[BaseUnit]], throw a *TypeError* exception.
-          1. If _value_ is *NaN*, return *NaN*.
+          1. If _value_ is ~not-a-number~, return ~not-a-number~.
+          1. If _value_ is ~positive-infinity~ or _value_ is ~negative-infinity~, return _value_.
```
gibson042, line 286: [Editorial Conventions: Comparisons](https://github.com/tc39/ecma262/wiki/Editorial-Conventions#comparisons:~:text=when%20comparing%20for%20equality%20against%20multiple%20options)
```suggestion
          1. If _intlMV_.[[Value]] is one of ~not-a-number~, ~positive-infinity~, or ~negative-infinity~, return _intlMV_.
```

spec.emu
```diff
@@ -268,32 +268,35 @@ location: https://tc39.es/proposal-amount/
...
           1. Let _targetOffset_ be _targetConv_.[[Offset]].
           1. If _sourceOffset_ is _targetOffset_, then
-            1. NOTE: This preserves a _value_ of *-0*<sub>𝔽</sub>.
-            1. Return _value_ × 𝔽(_sourceFactor_ / _targetFactor_).
-          1. Return _value_ × 𝔽(_sourceFactor_ / _targetFactor_) + 𝔽((_sourceOffset_ - _targetOffset_) / _targetFactor_).
+            1. NOTE: With equal offsets the conversion is purely multiplicative. Since unit factors are positive, this preserves a _value_ of ~negative-zero~.
+            1. If _value_ is ~negative-zero~, return ~negative-zero~.
+            1. Return _value_ × _sourceFactor_ / _targetFactor_.
+          1. If _value_ is ~negative-zero~, let _scalar_ be 0; otherwise, let _scalar_ be _value_.
+          1. Return _scalar_ × _sourceFactor_ / _targetFactor_ + (_sourceOffset_ - _targetOffset_) / _targetFactor_.
```
gibson042, line 296:
```suggestion
            1. If _intlMV_.[[Value]] is ~negative-zero~, return _intlMV_.
            1. Let _result_ be _intlMV_.[[Value]] × _sourceFactor_ / _targetFactor_.
          1. Else,
            1. If _intlMV_.[[Value]] is ~negative-zero~, let _scalar_ be 0; otherwise, let _scalar_ be _value_.
            1. Let result be _scalar_ × _sourceFactor_ / _targetFactor_ + (_sourceOffset_ - _targetOffset_) / _targetFactor_.
          1. Return the Intl MV Record { [[Value]]: _result_, [[StringDigitCount]]: 0, [[StringFractionDigitCount]]: 0 }.
```

spec.emu
```diff
@@ -446,23 +449,25 @@ location: https://tc39.es/proposal-amount/
...
-      1. If _v_ is a Number, then
-        1. Let _sourceValue_ be _v_.
-      1. Else if _v_ is a BigInt, then
-        1. Let _sourceValue_ be 𝔽(ℝ(_v_)).
-      1. Else,
-        1. Assert: _v_ is a String.
-        1. Let _sourceValue_ be StringToNumber(_v_).
+      1. Let _sourceMV_ be ! ToIntlMathematicalValue(_v_).
+      1. Let _sourceValue_ be _sourceMV_.[[Value]].
       1. Let _convertedValue_ be ? ConvertUnitValue(_sourceValue_, _sourceUnit_, _targetUnit_).
```
gibson042, line 454:
```suggestion
      1. Let _sourceMV_ be ! ToIntlMathematicalValue(_v_).
      1. Let _converted_ be ? ConvertUnitValue(_sourceMV_, _sourceUnit_, _targetUnit_).
      1. Let _convertedValue_ be _converted_.[[Value]].
```

spec.emu
```diff
@@ -446,23 +449,25 @@ location: https://tc39.es/proposal-amount/
...
       1. Let _result_ be OrdinaryObjectCreate(%Amount.prototype%, « [[AmountValue]], [[Unit]] »).
-      1. If _convertedValue_ is finite and (_fractionDigits_ is not *undefined* or _significantDigits_ is not *undefined*), then
+      1. If (_convertedValue_ is a mathematical value or _convertedValue_ is ~negative-zero~) and (_fractionDigits_ is not *undefined* or _significantDigits_ is not *undefined*), then
```
gibson042, line 456:
```suggestion
      1. If _fractionDigits_ is not *undefined* or _significantDigits_ is not *undefined*, let _toExplicitPrecision_ be *true*; else let _toExplicitPrecision_ be *false*.
      1. If _toExplicitPrecision_ is *true* and _convertedValue_ is either a mathematical value or ~negative-zero~, then
```

gibson042 (approved):
I thought there was too much resistance to mathematical-value multiplication and division, but am happy if that's not the case.

sffc (changes requested):
With #115, not all results become exact; for example, 0.3𝔽 * 3 still equals 0.8999999999999999𝔽. I think the way to specify this is to use `Math.fma`.

EDIT: This example is describing what the behavior _should be_, which is not what this PR specifies.

gibson042:
> With #115, not all results become exact; for example, 0.3𝔽 * 3 still equals 0.8999999999999999𝔽. I think the way to specify this is to use `Math.fma`.

Where in the changes of this PR do you see any step that could result in that calculation?

sffc:
> > With #115, not all results become exact; for example, 0.3𝔽 * 3 still equals 0.8999999999999999𝔽. I think the way to specify this is to use `Math.fma`.
> 
> Where in the changes of this PR do you see any step that could result in that calculation?

This PR appears to do all math in MV space, including converting a float to an Intl MV, which treats the float as the shortest decimal, not the full-precision floating point MV. However, that cannot be _implemented_ without decimal arithmetic, which is a hard no from engines. The approach in #115 is a compromise that gets us closer to exact arithmetic by removing the double-rounding in `a * b / c` but still allowing it to be done using floats.

eemeli (changes requested):
This seems like quite a surprising change, and would require explicit implementer approval before we apply it. Introducing a requirement for an implementation to be able to do decimal multiplication is... a lot. And rather like with `Math.fma()`, it's a capability that, if approved, should not only be available in this one corner of the spec.

So if you would like us to do this, I think you need to first present a separate proposal introducing that capability to the language, and then we use it here.

jessealama:
The thinking behind these changes was to imagine a world where we had `Math.fma` so the Number-based calculations offered by Amount would be, effectively, as good as it gets. The idea is that moving (more of) the calculations to mathematical values, as done here, captures that intent without actually using `Math.fma` in the spec text. But it sounds like what I'm up to here is going beyond what `Math.fma` can offer. At a minimum, an argument needs to be made for that proposition. I'll keep this PR open but mark it as a draft, for now, and will tackle the rounding issues in a separate PR using `Math.fma`. If that PR looks good, we can close this one.

sffc:
> But it sounds like what I'm up to here is going beyond what `Math.fma` can offer.

Right. We can't require implementors to do more than `Math.fma` can offer.

sffc:
Let me qualify that comment: The objective of FMA is to avoid multi-rounding of floats. It might be possible to specify _that_ behavior without using Math.fma, so long as it is done carefully. 

See some of my more recent comments in #115.

PR #128, opened 2026-07-02
Reject units that contain whitespace

### LINKED ISSUE #127: Should whitespace in unit strings be trimmed?
Looking ahead to use cases where a unit is provided dynamically, e.g. extracted from external data such as a web page, rather than hardcoded by a JS programmer, I wonder if want to consider normalizing units.

Today, trying to convert a `new Amount(42, " inch")` to centimeters would throw because CLDR's units.xml has no `" inch"`. This is understandable, but perhaps too strict.

One idea would be to trim the unit string on the left and the right before storing it. (Note that, as of today, we throw a `RangeError` if the empty string is supplied as a unit. One could view this as the extreme case of trimming a string that contains nothing but whitespace.)

Another idea would be to introduce a "strict mode" option in the constructor.

If we change the constructor in any of these ways, arguably `.convertTo` should be analogously updated.

jessealama:
In #127, the consensus is to reject unit strings containing whitespace. Here, we modify `GetAmountOptions` to throw a `RangeError` if given unit string that contains any code point in the union of the `WhiteSpace` or `LineTerminator` productions (taking inspiration from the definition of white space used by `TrimString` in 262). This follows the precedent set by Temporal and Intl.Locale.

Update the README to match.

Remove the `GetAmountConvertToOptions` AO, which has become identical to `GetAmountOptions`.

Closes #127

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Note: ⚠️  All property/method names up for bikeshedding.
 
 * `new Amount(value[, options])`. Constructs an Amount with the numerical value of `value`
   and optional `options`, of which the following are supported (all being optional):
-  * `unit` (String): A unit identifier associated with the numerical value, which must not be an empty string.
+  * `unit` (String): A unit identifier associated with the numerical value, which must not be an empty string and must not contain whitespace. Unit strings are never trimmed or otherwise normalized.
     As a shorthand, `options` may be given directly as a String, which is equivalent to passing `{ unit: options }`,
     so `new Amount(42, "meter")` is the same as `new Amount(42, { unit: "meter" })`.
   * `fractionDigits`: the number of fractional digits the mathematical value should have (can be less than, equal to, or greater than the actual number of fractional digits that the underlying mathematical value has when rendered as a decimal digit string)
--- a/spec.emu
+++ b/spec.emu
@@ -81,44 +81,7 @@ location: https://tc39.es/proposal-amount/
         </h1>
         <dl class="header">
           <dt>description</dt>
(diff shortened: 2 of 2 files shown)


eemeli (changes requested):
Let's be more strict here, and limit further the set of valid unit identifiers.

Also, I'm pretty sure that GetAmountConvertToOptions is an exact copy of GetAmountOptions?

jessealama:
> Let's be more strict here, and limit further the set of valid unit identifiers.

Perhaps ASCII alphanumeric only?

> Also, I'm pretty sure that GetAmountConvertToOptions is an exact copy of GetAmountOptions?

Ah, good spot. There was a time when we had _slightly_ different shapes for those two, reflected here, but I think we settled on making them the same.

gibson042 (changes requested):
The set of characters that must be excluded is far broader than just whitespace, especially if the unit will appear as-is in `toString()` serialization as is currently the case since #102. But limiting to ASCII is too excessive, because that would rule out reasonable units like "µs". I continue to believe that the right constraint is a minimal expansion of [UTS #35 Unit Identifiers](https://unicode.org/reports/tr35/tr35-general.html#Unit_Identifiers) that incorporates [IdentifierPartChar](https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#prod-IdentifierPartChar), e.g. matched by ECMAScript regular expression `/^[\p{ID_Continue}$]+(-[\p{ID_Continue}$]+)*$/u` (_note that UTS #35 **already** requires supporting counterintuitive unit identifers like "100", but we shouldn't expand further to include e.g. "-", "-100", or "foo--bar"_).

jessealama:
Superseded by #130

PR #137, opened 2026-07-12
Normative: Use the correct decimal digit count for non-decimal inputs

### LINKED ISSUE #135: Constructor parses non-decimal Strings but are stored without precision
Our string grammar uses `|StrNumericLiteral|`, so non-decimal strings like `"0x1"` do parse successfully in the Amount constructor. But the `StringIntlMV` AO gives such literals a `[[StringDigitCount]]` of 0, and the constructor passes that to `RenderInExponentialNotation`. But that AO requires a positive number of significant digits. The spec text seems to be broken for these kinds of inputs.

gibson042:
Fixes #135

### DIFF
--- a/spec.emu
+++ b/spec.emu
@@ -342,11 +342,11 @@ location: https://tc39.es/proposal-amount/
             1. Set _value_ to *+∞*<sub>𝔽</sub>.
           1. Else if _mv_ is ~negative-infinity~, then
             1. Set _value_ to *-∞*<sub>𝔽</sub>.
+          1. Else if _mv_ is ~negative-zero~, then
+            1. Set _value_ to RenderInExponentialNotation(_mv_, 1).
           1. Else,
-            1. If _parsed_ contains a |NonDecimalIntegerLiteral| Parse Node, then
-              1. Let _digitCount_ be the count of digits in the decimal representation of _intlMV_.[[Value]].
-            1. Else,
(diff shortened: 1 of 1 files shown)


spec.emu
```diff
@@ -343,7 +343,12 @@ location: https://tc39.es/proposal-amount/
           1. Else if _mv_ is ~negative-infinity~, then
             1. Set _value_ to *-∞*<sub>𝔽</sub>.
           1. Else,
-            1. Set _value_ to RenderInExponentialNotation(_mv_, _intlMV_.[[StringDigitCount]]).
+            1. If _parsed_ contains a |NonDecimalIntegerLiteral| Parse Node, then
+              1. Let _digitCount_ be the count of digits in the decimal representation of _intlMV_.[[Value]].
+            1. Else,
+              1. Let _digitCount_ be _intlMV_.[[StringDigitCount]].
+            1. Assert: _digitCount_ ≥ 1.
```
eemeli, line 350: This assertion fails if _value_ is the empty string. I think basing the conditional above on _intlMV_.[[StringDigitCount]] being 0 might work as intended?

eemeli (changes requested):

spec.emu
gibson042: Yeah, good catch. Fixed.

jessealama (approved):

Issue #140, opened 2026-08-06
Precision of conversion between non-primitive units


eemeli:
Our work on unit conversion in Amount has thus far focused on single unit conversions, such as converting between units like `foot` and `meter`. For these, the total count of units is sufficiently limited that we ought to be able to precalculate the values necessary to ensure that conversions are as accurate as possible when done in the Number space.

However, we also intend to cover unit categories like "speed", where it's reasonable to expect that a user might want to convert e.g. a speed measured in `meter-per-second` to `mile-per-hour`. The conversion here is `x / (0.3048 * 5280) * 3600`, and even if an implementation were to precalculate and store the meters-to-miles multiplier (`1 / (0.3048 * 5280)`), we'd still end up with two sequential multiplications, with an intermediate result potentially enforced to be a Number.

While we could precalculate all of `1 / (0.3048 * 5280) * 3600`, that doesn't scale well to covering all possible combinations, which would include all sorts of unlikely but not impossible conversion pairs like going from `watt-per-joule` to `horsepower-per-calorie`.

I suggest that it would be reasonable for us not to promise the same kind of accuracy with `-per-` units that we can promise for single-unit conversions. As this does run the risk of allowing for imprecise outputs, we should consider limiting the `roundingMode` values that we offer to minimize the impact of these imprecisions.

jessealama:
In the [JS Numerics call](https://docs.google.com/document/d/1O2EQC61TIDtkcvDSkhDf4N_R9GioT0foU2tH9HBdMdQ/) on July 30th it looks like there a few other considerations were raised in connection with this issue:

* Should we define a blessed set of units for which we _do_ provide some kind of precision guarantee (e.g., combinations of SI units)?
* Whatever decision we make, it seems preferable to have test262 tests for cases where we expect precision loss rather than let such tricky calculations be implementation-defined.

sffc:
I'd like to see how close we get to running out of room in an i64 rational for the conversation rate precalculation, which we'd then feed to FMA.

Issue #118, opened 2026-05-21
Amount + Sequence Units


sffc:
Now that both of these proposals are Stage 2 🎊, we need to figure out how they fit together.

We had some brainstorming discussions with myself and @eemeli and @lucacasonato and @ljharb and others at TC39.

Some points that people have raised; not all of these bullet points are universally held:

- Adding sequence units should not (significantly) hurt ergonomics of non-sequence units
- Sequence units should be first-class in most functionality, including the constructor, convertTo, and formatting
- We should try to keep just 1 Amount type, not adding a second to support sequence units

There are a handful of shapes to consider:

1. Amount is a single unit; sequence units are returned as arrays
2. Amount is a sequence unit; single units are returned as objects
3. Amount is a sequence unit; getting a single unit means querying the first entry of it

A design that seems to solve _most_ of the design constraints is an `Amount.prototype.total` function that behaves like this:

```javascript
let amt1 = new Amount(123, { unit: "centimeter" });
amt1.getMeasures()  // [{ unit: "centimeter", value: 123 }]
amt1.total()  // { unit: "centimeter", value: 123 }]

let amt2 = amt1.convertTo({ units: ["meter", "centimeter"] });
amt1.getMeasures()  // [{ unit: "meter", value: 1 }, { unit: "centimeter", value: 23 }]
amt2.total()  // { unit: "meter", value: 1.23 }
```

Some open questions:

- Maybe `.getMeasures()` should be a getter that returns a deeply frozen array
- Maybe we should allow `unit: "meter-and-centimeter"` and/or drop the `units` option
- Should we still have a `.unit` getter, and if so, what should it return?
- Should we still have a `.value` getter, and if so, what should it return?
- Should we allow only sanctioned units to be a member of a sequence unit, so that we can enforce the correct order and implement `.total()`, or should we throw in `.total()` when the unit is not sanctioned? Also see https://github.com/tc39/proposal-amount/issues/117

gibson042:
> A design that seems to solve _most_ of the design constraints is an `Amount.prototype.total` function that behaves like this:
> 
> let amt1 = new Amount(123, { unit: "centimeter" });
> amt1.getMeasures()  // [{ unit: "centimeter", value: 123 }]
> amt1.total()  // { unit: "centimeter", value: 123 }]
> 
> let amt2 = amt1.convertTo({ units: ["meter", "centimeter"] });
[...]

Issue #116, opened 2026-05-21
Amount toString serialization discussions


sffc:
Diving straight in: there are a number of constraints around toString, which could carry different weight:

- Should be consistently (always or never) parseable by existing parse functions like JSON.parse and Number.parseFloat
- Should work seamlessly with sequence units
- Should be parseable by a custom parser
- Should be developer-readable
- Should build on top of existing serialization formats if possible

Any other constraints? Please comment below.

With that in mind, here are some possible syntaxes for these inputs:

- Unitless: { value: 1, unit: null }
- Meter: { value: 5, unit: "meter" }
- Sequence: { value: { pound: 7, ounce: 1 }, unit: "pound-and-ounce" }

| Name | Unitless | Meter | Sequence |
|---|---|---|---|
| Stage 2 | [1e+0 ~] | [5e+0 meter] | [7e+0 pound, 1e+0 ounce] |
| With IXDTF annotations | 1e+0[] | 5e+0[meter] | 7e+0,1e+0[pound-and-ounce] |
| At-sign IXDTF | `@1e+0` | `@5e+0[meter]` | `@7,1e+0[pound-and-ounce]` |
| Raw JSON | {"value":1e+0,"unit":null} | {"value":5e+0,"unit":"meter"} | {"value":{"pound":7e+0,"ounce":1e+0},"unit":"pound-and-ounce"} |
| Nicer JS, A | [{value:1e+0}] | [{value:5e+0,unit:"meter"}] | [{value:7e+0,unit:"pound"},{value:1e+0,unit:"ounce"}] |
| Nicer JS, B | [1e+0] | [{meter:5e+0}] | [{pound:7e+0},{ounce:1e+0}] |
| Inspired by ISO 8601 P-string | A1e+0 | A5e+0meter | A7e+0pound1e+0ounce |
| P-string with well-known abbreviations | A1e+0 | A5e+0M | A7e+0L1e+0O |
| P-string with annotations | A1e+0 | A5e+0[meter] | A7e+0[pound]1e+0[ounce] |

Please feel free to offer more ideas.

gibson042:
[...]

PR #141, opened 2026-08-08
Add sequence units to explainer

### LINKED ISSUE #143: What does `significantDigits` mean for a sequence unit?
Splitting off from #141: How should we handle

```js
new Amount([5, 11.1], { unit: "foot-and-inch", significantDigits: 3 })
```

?

I think one can hold one of two different views of sequence units, and we need to decide which one is in effect in thinking about significant digits.

On the "array view", the option is ambiguous: which entry does it apply to? `fractionDigits` has an answer to this question: the fractional part lives entirely inside the smallest unit. Significance is counted from the leading digit, which sits in the first entry, so counting across entries means counting in a mixed radix, not necessarily guaranteed to be multiples of ten (such as foot-and-inch). It makes sense to throw.

On the "scalar view", [5, 11.1] foot-and-inch just is 71.1 inches written more readably, and that can be expressed with (exactly!) three significant digits. On this view, there's a canonical unit (here, inches). We can easily do arithmetic (multiplying by integers and summing) and get a definite answer.

(Canonicalization generates yet another question: what do we store? Should we store the canonicalized scalar "7.11e+0", thereby turning the input array into a primitive, or should we re-represent the canonicalized value as an array (i.e., `[5, 11]`)?

(Whatever we decide presumably impacts `.convertTo`, too.)

eemeli:
This adds what I understand to be our current consensus on how to support sequence units (see e.g. #118 and [proposal-intl-sequence-units](https://github.com/tc39/proposal-intl-sequence-units)), i.e. using an array as the `.value` where appropriate.

The spec is not (yet) updated by this change, just the explainer.

Going through the details, there are a few cases that I'm not sure we've covered previously, and on some of which this PR is opinionated:

- Combining `significantDigits` with sequence units — I recommend we throw an error for this, as the general solution here would effectively require dealing with non-decimal bases.
- Type of not-smallest unit values when applying `fractionDigits` — Are all array values stringified? For example, with
  ```js
  let a = new Amount([5, 11.1], { fractionDigits: 0 })
  a.value  //  [5, '1.1e+1']  or  ['5e+0', '1.1e+1']
  ```
  both options seem somewhat reasonable to me.
- The non-localized serialization of not-smallest unit values — we require these to always be integers, so shouldn't we also serialize them as plain integers? Without `significantDigits`, these can't have precision information attached to them.
- As with the `-per-` units mentioned in [#140](https://github.com/tc39/proposal-amount/issues/140), conversion to/from `-and-` units requires multiple sequential Number operations that can introduce deviations compared to decimal math.

Ping @sffc, @gibson042, @ljharb

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,9 +46,8 @@ We propose creating a new `Amount` primordial containing an immutable numeric va
 
 Amount will have the following read-only properties:
 
-Note: ⚠️  All property/method names up for bikeshedding.
-
-* `value` (Number or BigInt or String): The numerical value of the amount.
+* `value: number | bigint | string | ReadonlyArray<number | bigint | string>` —
+  The numerical value of the amount.
   The type of the value used in the constructor is retained,
   except that any non-finite value is a Number (`Infinity`, `-Infinity`, or `NaN`)
   and any value that was potentially affected by precision options is a String.
@@ -57,14 +56,21 @@ Note: ⚠️  All property/method names up for bikeshedding.
   (decimal exponential notation with an explicitly signed exponent
   and a significand that is either exactly 0 or has a positive absolute value less than 10)
   but not necessarily limited by the same bounds.
-* `unit` (String or not defined): The unit of measurement associated with the Amount's numerical value.
+
+  An Array `value` is used when the `unit` includes at least one `-and-` substring.
+  These represent [sequence units], such as "5 feet, 11 inches".
+  The Array length matches the number of `-and-` separated parts of the `unit`,
+  and all except for the last Array entry always represent integers.
+
+* `unit: string | undefined` — The unit of measurement associated with the Amount's numerical value.
   An undefined value indicates "no unit supplied".
 
 [Number.p.toExponential]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toExponential
+[sequence units]: https://github.com/tc39/proposal-intl-sequence-units
 
 ### Constructor
 
-* `new Amount(value[, options])`. Constructs an Amount with the numerical value of `value`
+* `new Amount(value[, options])`. Constructs an Amount with `value`
   and optional `options`, of which the following are supported (all being optional):
   * `unit` (String): A unit identifier associated with the numerical value.
     A unit identifier is one or more nonempty segments separated by single hyphens (`-`),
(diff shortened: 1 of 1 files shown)


jessealama (approved):
Thanks! I agree that trying to specify significant digits when working with sequence units should throw, at least for now. I think it's worth discussing some of your other opinionated bits in new issues.

README.md
```diff
@@ -74,20 +80,36 @@ Note: ⚠️  All property/method names up for bikeshedding.
...
+  and all except for the last Array entry must represent an integer;
+  otherwise a TypeError is thrown.
+
   If either `fractionDigits` or `significantDigits` is set,
   the `value` is rounded accordingly,
   and is stored as a String (if finite) or Number (if not finite).
+  If `value` is an Array and `fractionDigits` is set,
+  the rounding is only applied to the last value in the Array.
   If both `fractionDigits` and `significantDigits` are set, a RangeError is thrown.
+  If `value` is an Array and `significantDigits` is set, a RangeError is thrown.
```
gibson042: Here too (but leaving the current RangeError, which is already subject to #92).
```suggestion
  If both `fractionDigits` and `significantDigits` are set, a RangeError is thrown.
  If `value` is an Array and `fractionDigits` is set,
  the rounding is only applied to the last value in the Array.
  If `value` is an Array and `significantDigits` is set, a TypeError is thrown.
```

README.md
```diff
@@ -116,17 +138,30 @@ The object prototype would provide the following methods:
   the result will be rounded according to the precision options,
   and the returned Amount will have a String `value`.
   Otherwise, the returned Amount will have a Number `value`.
+  If the conversion target is a sequence unit and `fractionDigits` is set,
+  the rounding is only applied to the last value in the Array.
   If both `fractionDigits` and `significantDigits` are set, a RangeError is thrown.
+  If the conversion target is a sequence unit and `significantDigits` is set, a RangeError is thrown.
```
gibson042: And here.
```suggestion
  If both `fractionDigits` and `significantDigits` are set, a RangeError is thrown.
  If the conversion target is a sequence unit and `fractionDigits` is set,
  the rounding is only applied to the last value in the Array.
  If the conversion target is a sequence unit and `significantDigits` is set, a TypeError is thrown.
```

gibson042 (approved):
I agree with rejecting `significantDigits` for sequence units. No strong opinions on the other questions right now.

README.md
```diff
@@ -116,17 +138,30 @@ The object prototype would provide the following methods:
...
+  In other words, conversion to `foot-and-inch` is valid, but
+  `inch-and-foot`, `foot-and-centimeter`, and `foot-and-gallon` are not valid conversion targets.
+
 * `toString()`: A string representation of the Amount.
   Returns a digit string together with the unit, surrounded by square brackets (for example, `"[1.23+e0 kilogram]"`).
   If the Amount does not have a unit,
   the tilde `~` (U+007E) is used in place of the unit (for example, `"[4.2+e1 ~]"`).
+  If the Amount has a sequence unit, the unit identifier is split to its component units,
+  and the component unit and their values are joined by `", "`
+  (for example, `"[6e+1 degree, 1.1e+1 arc-minute, 3.141e+1 arc-second]"`).
```
sffc, line 163: I think we should iterate on the string format more but we can do that in an issiue

README.md
```diff
@@ -160,6 +195,18 @@ For example, to convert 1.75 feet to inches, the following mathematical operatio
...
+each of the individual parts of the unit are converted separately,
+and the Number results then summed together.
+
+When converting to a sequence unit,
+the source unit value or values are first converted to the least-magnitude unit of the target sequence unit.
+Euclidean division is then applied to this Number using the integer multiplier
+from the smaller-magnitude unit to the next unit by magnitude.
+The remainder is assigned to the smaller-magnitude unit,
+and the quotient to the larger-magnitude unit.
+If the sequence unit contains more than two units, this process is repeated.
```
sffc, line 208: I'm not sure if Euclidean division is correct here. We want -15 inches to be (-1ft, -3in), not (-2ft, 9in). We want Euclidean division for calendar arithmetic.

README.md
jessealama: We do have an issue for this: #116

README.md
jessealama: I've made an issue out of this (#156).

PR #134, opened 2026-07-10
Parse Amount string values with StrNumericLiteral, extended to accept "NaN"

jessealama:
The Amount constructor now parses String values with ECMA-262's `StrNumericLiteral` production, which this proposal extends with a `NaN` alternative and matching `StringNumericValue` and `StringIntlMV` rules. The string `"NaN"` is now accepted, while the empty string, whitespace-only strings (to which the runtime semantics of `StringNumericLiteral` assign the value 0), and strings with leading or trailing white space all throw a RangeError.

An earlier version of this PR defined a dedicated grammar for acceptable strings; review feedback simplified that to extending an existing ECMA-262 production.

Relates to #132. This change maintains support for `"NaN"` as an argument to the constructor, just making clear that we don't support `"NaN"` with leading or trailing whitespace. If we want to drop support for `"NaN"` (and `"Infinity"` and `"-Infinity"`), which I think is reasonable, we can tackle that in a different PR.

Closes #133

PR #146, opened 2026-08-13
Normative: Accept only Strings that denote finite values

### LINKED ISSUE #105: Should the constructor accept String versions of Infinity and NaN?
Non-finite values should always be represented as Number values, never strings.

We'll need to discuss separately whether `new Amount('Infinity')` should throw, or end up with `.value: Infinity`.

_Originally posted by @eemeli in https://github.com/tc39/proposal-amount/pull/104#discussion_r3166970673_

### LINKED ISSUE #132: Unreachable NaN branch in the constructor
The Amount constructor parses String arguments with `|StringNumericLiteral|` and throws a `RangeError` if parsing fails. The algorithm then contains a branch that checks whether the parsed mathematical value is `~not-a-number~` and, if so, stores the Number value `NaN`. The intention is clear enough, but that branch is actually unreachable: `|StringNumericLiteral|` does not match `"NaN"`. Thus, `new Amount("NaN")` throws a `RangeError` before checking for `~not-a-number~`.

Comparing with how Number arguments are handled in the constructor, notice the asymmetry:

```js
new Amount(NaN);      // Amount with value NaN
new Amount("NaN");    // RangeError
```

### LINKED ISSUE #133: Constructor accepts the empty string and whitespace-only strings
The constructor parses whitespace-only String values with `|StringNumericLiteral|` and assigns them the mathematical value 0:

```js
new Amount("");       // Amount with value 0
new Amount("   ");    // Amount with value 0
```

The Amount constructor throws a `RangeError` for other unparseable input (e.g. `"abc"`). Empty strings (or whitespace-only strings) quietly accepted as zero seems like an oversight.

jessealama:
The Amount constructor now parses String values with a new grammar production, `|StrFiniteNumericLiteral|`, defined as a `|StrNumericLiteral|` excluding the infinities. Because `|StrNumericLiteral|` does not allow whitespace, any whitespace-only strings or strings with leading or trailing whitespace, will fail to parse. `"NaN"` also fails to parse.

This nails down the direction discussed in https://github.com/tc39/proposal-amount/issues/105, https://github.com/tc39/proposal-amount/issues/132, and https://github.com/tc39/proposal-amount/issues/133, and supersedes https://github.com/tc39/proposal-amount/pull/134, keeping its grammar-based parsing and whitespace rejection while dropping the NaN grammar extension.


README.md
```diff
@@ -50,8 +50,9 @@ Note: ⚠️  All property/method names up for bikeshedding.
 
 * `value` (Number or BigInt or String): The numerical value of the amount.
   The type of the value used in the constructor is retained,
-  except that any non-finite value is a Number (`Infinity`, `-Infinity`, or `NaN`)
-  and any value that was potentially affected by precision options is a String.
+  except that any value that was potentially affected by precision options is a String.
+  Only a Number `value` can be non-finite (`Infinity`, `-Infinity`, or `NaN`);
+  String representations of non-finite values are rejected.
```
gibson042: Rejection is _behavior_ and therefore seems inappropriate for section "Properties".
```suggestion
  A Number `value` can be non-finite (`Infinity`, `-Infinity`, or `NaN`).
```

README.md
```diff
@@ -80,8 +81,11 @@ Note: ⚠️  All property/method names up for bikeshedding.
 
   Attempting to construct an Amount from a `value` that is not a Number or BigInt or String will throw a TypeError.
   When constructing an Amount from a String `value`,
-  its mathematical value is parsed using [StringNumericLiteral](https://tc39.es/ecma262/#prod-StringNumericLiteral)
-  or a RangeError is thrown.
+  the string must be exactly a numeric literal denoting a finite value:
+  a [StrNumericLiteral](https://tc39.es/ecma262/#prod-StrNumericLiteral)
+  other than `"Infinity"`, `"+Infinity"`, or `"-Infinity"`;
+  anything else (including `"NaN"`, the empty string,
+  and strings with leading or trailing white space) throws a RangeError.
```
gibson042: Grammar nit:
```suggestion
  other than `"Infinity"`, `"+Infinity"`, or `"-Infinity"`.
  Anything else (including `"NaN"`, the empty string,
  and strings with leading or trailing white space) throws a RangeError.
```

spec.emu
gibson042: Suggestion: refactor these steps to make clear that a String _value_ always ends up in exponential notation.
```patch
-            1. Set _value_ to RenderInExponentialNotation(_mv_, 1).
+            1. Set _digitCount_ to 1.
           1. Else,
             1. Assert: _mv_ is a mathematical value.
             1. If _intlMV_.[[StringDigitCount]] ≥ 1, let _digitCount_ be _intlMV_.[[StringDigitCount]]; else let _digitCount_ be the count of digits in the decimal representation of _mv_.
-            1. Assert: _digitCount_ ≥ 1.
-            1. Set _value_ to RenderInExponentialNotation(_mv_, _digitCount_).
+          1. Set _value_ to RenderInExponentialNotation(_mv_, _digitCount_).
```
```suggestion
            1. Set _digitCount_ to 1.
          1. Else,
            1. Assert: _mv_ is a mathematical value.
            1. If _intlMV_.[[StringDigitCount]] ≥ 1, let _digitCount_ be _intlMV_.[[StringDigitCount]]; else let _digitCount_ be the count of digits in the decimal representation of _mv_.
          1. Set _value_ to RenderInExponentialNotation(_mv_, _digitCount_).
```

gibson042 (approved):
This strikes me as a good approach.

gibson042 (approved):
This strikes me as a good approach.

gibson042 (approved):
This strikes me as a good approach.

spec.emu
```diff
@@ -321,19 +331,17 @@ location: https://tc39.es/proposal-amount/
...
-          1. Else if _mv_ is ~negative-zero~, then
+          1. If _mv_ is ~negative-zero~, then
             1. Set _value_ to RenderInExponentialNotation(_mv_, 1).
           1. Else,
             1. Assert: _mv_ is a mathematical value.
             1. If _intlMV_.[[StringDigitCount]] ≥ 1, let _digitCount_ be _intlMV_.[[StringDigitCount]]; else let _digitCount_ be the count of digits in the decimal representation of _mv_.
             1. Assert: _digitCount_ ≥ 1.
-            1. Set _value_ to RenderInExponentialNotation(_mv_, _digitCount_).
+            1. Set _digitCount_ to 1.
+          1. Else,
```
eemeli: This is a mistake, right? This is a second Else for the same If statement, and the "Set _digitCount_ to 1" is a bit surprising.

eemeli (changes requested):

spec.emu
```diff
@@ -321,19 +331,17 @@ location: https://tc39.es/proposal-amount/
...
+            1. Assert: _mv_ is a mathematical value.
+            1. If _intlMV_.[[StringDigitCount]] ≥ 1, let _digitCount_ be _intlMV_.[[StringDigitCount]]; else let _digitCount_ be the count of digits in the decimal representation of _mv_.
+          1. Set _value_ to RenderInExponentialNotation(_mv_, _digitCount_).
```
gibson042: Looks like GitHub applied the previous suggestion incorrectly.
```suggestion
          1. If _mv_ is ~negative-zero~, then
            1. Set _digitCount_ to 1.
          1. Else,
            1. Assert: _mv_ is a mathematical value.
            1. If _intlMV_.[[StringDigitCount]] ≥ 1, let _digitCount_ be _intlMV_.[[StringDigitCount]]; else let _digitCount_ be the count of digits in the decimal representation of _mv_.
          1. Set _value_ to RenderInExponentialNotation(_mv_, _digitCount_).
```

jessealama: I've taken care of this now, manually (I wasn't able to apply your suggestion in the web UI since it affects deleted lines).

spec.emu
jessealama: I think this has been resolved now thanks to @gibson042 's edit (which I had to manually apply).

spec.emu
```diff
@@ -321,19 +331,11 @@ location: https://tc39.es/proposal-amount/
...
-            1. Set _value_ to *NaN*.
-          1. Else if _mv_ is ~positive-infinity~, then
-            1. Set _value_ to *+∞*<sub>𝔽</sub>.
-          1. Else if _mv_ is ~negative-infinity~, then
-            1. Set _value_ to *-∞*<sub>𝔽</sub>.
-          1. Else if _mv_ is ~negative-zero~, then
-            1. Set _value_ to RenderInExponentialNotation(_mv_, 1).
-          1. Else,
+          1. Let _digitCount_ be 1.
+          1. If _mv_ is not ~negative-zero~, then
```
eemeli, line 335: This is the wrong conditional, as it's not accounting for `'-0.0'`, which should have a _digitCount_ of 2. The retention of the digit count for that is [specifically noted](https://tc39.es/proposal-intl-keep-trailing-zeros/#:~:text=The%20value%20of%20digitCount%20is%20retained) in keep-trailing-zeros.

Should drop this check, and only fall back to "the count of digits in the decimal representation of _mv_" if _intlMV_.[[StringDigitCount]] is 0, for which we can assert that "_mv_ is a mathematical value".

Also, not sure about this, but do we want to specify "decimal integer representation" as we can only end up on that path if the input was a NonDecimalIntegerLiteral, right? Or maybe strengthen the assertion to something like "_mv_ is an integer mathematical value"?

eemeli (changes requested):

spec.emu
jessealama: Good spot. I've taken care of that. I like asserting that the MV is an integer.

spec.emu
```diff
@@ -321,19 +331,12 @@ location: https://tc39.es/proposal-amount/
...
-          1. Else if _mv_ is ~negative-zero~, then
-            1. Set _value_ to RenderInExponentialNotation(_mv_, 1).
+          1. If _intlMV_.[[StringDigitCount]] ≥ 1, then
+            1. Let _digitCount_ be _intlMV_.[[StringDigitCount]].
           1. Else,
-            1. Assert: _mv_ is a mathematical value.
-            1. If _intlMV_.[[StringDigitCount]] ≥ 1, let _digitCount_ be _intlMV_.[[StringDigitCount]]; else let _digitCount_ be the count of digits in the decimal representation of _mv_.
-            1. Assert: _digitCount_ ≥ 1.
-            1. Set _value_ to RenderInExponentialNotation(_mv_, _digitCount_).
+            1. Assert: _mv_ is a non-negative integer.
```
eemeli: How/where did the "non-negative" assertion creep in here? It's completely wrong.
```suggestion
            1. Assert: _mv_ is an integer.
```

eemeli (changes requested):

spec.emu
jessealama: Ack, quite right. Fixed.

eemeli (approved):
Looks good with these changes. @gibson042, PTAL?

README.md
```diff
@@ -49,8 +49,8 @@ Amount will have the following read-only properties:
 * `value: number | bigint | string | ReadonlyArray<number | bigint | string>` —
   The numerical value of the amount.
   The type of the value used in the constructor is retained,
-  except that any non-finite value is a Number (`Infinity`, `-Infinity`, or `NaN`)
-  and any value that was potentially affected by precision options is a String.
+  except that any value that was potentially affected by precision options is a String.
+  A Number `value` can be non-finite (`Infinity`, `-Infinity`, or `NaN`).
```
gibson042: Suggestion for better indicating confinement of non-finite values to the Number type:
```suggestion
  except that any value that was potentially affected by precision options is a String.
  A non-finite value is always represented as a Number (`Infinity`, `-Infinity`, or `NaN`).
```

spec.emu
```diff
@@ -351,6 +354,9 @@ location: https://tc39.es/proposal-amount/
       <emu-note>
         <p>When no precision options are given, Number and BigInt arguments are stored directly in [[AmountValue]], preserving the original type. String arguments and values resulting from precision options are stored as a String in canonical exponential notation, preserving the precision implied by the input.</p>
       </emu-note>
+      <emu-note>
+        <p>A String argument must be exactly a numeric literal denoting a finite value. *"NaN"*, *"Infinity"*, *"+Infinity"*, and *"-Infinity"* are rejected, as well as Strings with leading or trailing white space, even though the corresponding Number values are accepted. *""* is also rejected.</p>
```
gibson042: Wording suggestions:
```suggestion
        <p>A String argument must denote a finite mathematical value. *"NaN"*, *"Infinity"*, *"+Infinity"*, *"-Infinity"*, and *""* are rejected, as are Strings with leading or trailing white space or other extraneous contents.</p>
```

gibson042 (approved):
LGTM

PR #161, opened 2026-09-10
Specify sequence units

jessealama:
This PR adds spec text for sequence units. The unit identifier grammar gains a `SequenceUnitIdentifier` production and reserves the segment `"and"`, so the components of a unit identifier are found by splitting it at `"-and-"`. Amounts with a sequence unit store one entry per component as a List in `[[AmountValue]]` and return a fresh frozen Array from the `value` getter. The constructor accepts either an Array or an object with one property per component; every entry but the last must be an integer, and no entry may be negative while another is positive. `toString` renders each entry with its component unit. 

`convertTo` needed the biggest update. A sequence source is validated and totalled exactly in its smallest component before converting, and the source is now checked before the options are read for every kind of source. A sequence target is produced by converting to the smallest component, rounding there, and splitting with truncating division so that every entry has the sign of the source, negative zero included. Convertibility of a sequence is decided from the CLDR conversion factors. `toLocaleString` formats through proposal-intl-sequence-units' `FormatNumericSequence`.

This work builds on convert-options (#110) because we need `SelectTargetUnit` and the unit category data.

This PR also refactors `Number.prototype.toExponential` in place. Its rendering steps move into a shared `RenderInExponentialNotation` abstract operation, which Amount uses for every value it renders in exponential notation, so the two cannot drift apart. As part of that, a Number value is now rendered with the fewest significant digits that identify it, so an Amount holding the Number 100 renders as `"1e+2"` rather than `"1.00e+2"`.

Rounding under precision options no longer goes through an Intl formatter. `RoundToFractionDigits` and `RoundToSignificantDigits` round a mathematical value directly. To support them, `GetUnsignedRoundingMode`, its table of unsigned rounding modes, and `ApplyUnsignedRoundingMode` are promoted unchanged from ECMA-402 to the Mathematical Operations clause of ECMA-262, and the ECMA-402 amendments delete them from their old home. They are pure arithmetic and the canonical definitions of the nine rounding modes. `CreateFormatterObject` and the detour through `FormatNumericToString` and its formatted string are gone.

The README is updated, too.


eemeli (changes requested):
> This PR also refactors `Number.prototype.toExponential` in place. [...]
> 
> [...] To support them, `GetUnsignedRoundingMode`, its table of unsigned rounding modes, and `ApplyUnsignedRoundingMode` are promoted unchanged from ECMA-402 to the Mathematical Operations clause of ECMA-262, and the ECMA-402 amendments delete them from their old home.

Could these changes be spun off into a separate PR, with the sequence unit changes building on top of it? That would make review quite a bit easier.

That PR doesn't need to be editorial only, just the parts of this that aren't about sequence units specifically.

jessealama:
> > This PR also refactors `Number.prototype.toExponential` in place. [...]
> > [...] To support them, `GetUnsignedRoundingMode`, its table of unsigned rounding modes, and `ApplyUnsignedRoundingMode` are promoted unchanged from ECMA-402 to the Mathematical Operations clause of ECMA-262, and the ECMA-402 amendments delete them from their old home.
> 
> Could these changes be spun off into a separate PR, with the sequence unit changes building on top of it? That would make review quite a bit easier.
> 
> That PR doesn't need to be editorial only, just the parts of this that aren't about sequence units specifically.

Sounds good. I'll break up the arithmetic stuff into a separate PR.

README.md
```diff
@@ -98,13 +99,16 @@ Amount will have the following read-only properties:
...
+  The Object is either an Array with exactly one entry per `-and-`-separated part of the `unit`
+  (any other length throws a TypeError),
+  or an object with one property named after each part, such as `{ foot: 5, inch: 11 }`
+  (a missing property throws a TypeError; extra properties are ignored).
   These represent [sequence units], such as "5 feet, 11 inches".
-  The Array length must match the number of `-and-` separated parts of the `unit`,
-  and all except for the last Array entry must represent an integer;
-  otherwise a TypeError is thrown.
+  All except the last entry must represent an integer,
+  and no entry may be negative while another is positive;
```
gibson042, line 110: How about this?
```suggestion
  and all non-zero entries must have the same sign;
```

README.md
```diff
@@ -157,11 +161,12 @@ The object prototype would provide the following methods:
   or if the resolved conversion target is not valid for the Amount's unit
   (such as attempting to convert a mass unit into a length unit).
 
-  If the conversion target is a sequence unit,
-  a TypeError is thrown if the individual units are not convertible with each other
-  using integer multipliers, with decreasing magnitude.
-  In other words, conversion to `foot-and-inch` is valid, but
-  `inch-and-foot`, `foot-and-centimeter`, and `foot-and-gallon` are not valid conversion targets.
+  If either the source or the target is a sequence unit,
+  a TypeError is thrown unless its component units share a CLDR base unit without an offset,
```
gibson042, line 165: Why require "without an offset"?

spec.emu
```diff
@@ -461,14 +798,20 @@ location: https://tc39.es/proposal-amount/
...
     <h1>get Amount.prototype.value</h1>
-    <p>This accessor property, whose set accessor function is *undefined*, returns the numeric value of the Amount. Its get accessor function performs the following steps when called:</p>
+    <p>This accessor property, whose set accessor function is *undefined*, returns the numeric value of the Amount, or a frozen Array of the values of a sequence-unit Amount. Its get accessor function performs the following steps when called:</p>
     <emu-alg>
       1. Let _O_ be the *this* value.
       1. Perform ? RequireInternalSlot(_O_, [[AmountValue]]).
-      1. Return _O_.[[AmountValue]].
+      1. Let _v_ be _O_.[[AmountValue]].
+      1. If _v_ is a List, then
+        1. Let _array_ be CreateArrayFromList(_v_).
```
gibson042, line 807: Returning fresh objects from a getter makes for bad DX—`amount.value !== amount.value` is just really unpleasant behavior. I'd _much_ prefer always returning the same frozen array.

gibson042 (commented):
Some comments on the approach.

jessealama:
Marking this as a draft while we work on splitting this up.

